PR #55, opened 2021-04-19
changed the energy detector to use a proxy for the output

### LINKED ISSUE #52: [Issue] Energy Detector sometimes voids energy, or is an infinit energy source without input
# Describe

## Expected

- measures the throughput and allows to limit it
- input and output of the Energy Detector are always the same
- no energy is lost (or only a bit as energy consumption of the Energy Detector)

## Observed

1. one time it suddenly was an infinity energy source even without an input attached (it gave 1.6 kFE/t)
2. now it drains any available energy on the input and voids the excess
    - multiple in parallel share the input evenly
    - limiting the throughput works, but it is used up to max even without an attached cable
3. two of our Energy Detectors are limited to 0 and don't have an input source attached, but still report a throughput of last time it was transmitting power

# Reproduce

1. use an not infinite energy source (Bigreactors - Basic Reactor)

## used mods
- Bigreactors - Basic Reactor
- Flux Networks
- Rftoolspower - Powercell (High)
- Makanism - Basic Universal Cable
- many others not directly interfacing with this (ask if you need to know more)

**Versions:**

- Forge version: 36.1.4
- AdvancedPeripherals version: 0.4.4b

bessw:
this should fix #52, the energy detector receives and forwards the energy now completely passiv instead of actively draining energy from the input.

## notes
I still get in lua sometimes miss matching results from EnergyDetector.getTransferRate() when using Mekanism universal cables, but this might be a Mekanism bug since it works now as expected when using other cables.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/EnergyDetectorPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/EnergyDetectorPeripheral.java
@@ -21,13 +21,13 @@ public class EnergyDetectorPeripheral extends BasePeripheral {
 
     @LuaFunction(mainThread = true)
     public final int getTransferRateLimit() {
-        return tileEntity.maxTransferRate;
+        return tileEntity.storage.getMaxTransferRate();
     }
 
     @LuaFunction(mainThread = true)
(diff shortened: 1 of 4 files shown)


Seniorendi:
This looks good. 
Thank you for your help. 👍🏻

Issue #58, opened 2021-04-22
[Feature] AR goggles


Olfi01:
Glasses for a player that can be adressed with a unique ID and show an overlay to the player which can be customized by a computer (like the terminal glasses in OpenPeripherals).
Using some kind of connection block that the glasses have been linked to and a computer next to it.
Also, it would be awesome to be able to put on the glasses in a specific "glasses" slot in the curios inventory, if curios is installed

Olfi01:
They've been implemented by #59 and are in 0.5b, so I guess this issue can be closed :D

PR #59, opened 2021-04-23
AR Goggles

Olfi01:
As suggested in #58 (by me), I implemented the AR goggles.

I'd be happy to provide documentation for the lua methods, too, I just don't know where I'd put that.

This makes the mod dependent on Curios, but I would consider that a worthwile tradeoff, even more so since Curios is already part of most modpacks.


Wojbie:
Why not make Curios optional dependency? Would that be possible? Make it act as helmet if Curios is not installed.

Or skip it all together and make googles work from anywhere in player inventory?

Seniorendi:
I completely agree to what Wojbie means.
And could you send a text to the player when they link the AR goggles?

Olfi01:
@Wojbie @Seniorendi 
Curios as an optional dependency doesn't work the way it is implemented right now, since I am using a few convenience methods offered by it. I suppose it would be possible to refactor the code in a way to make curios optional, but I would have to look into that later today or tomorrow.
As for the text when linking the AR goggles, sure, I will add that as well.

Olfi01:
Do NOT merge yet, still needs one fix

UnhappyCodings:
It would be cool if you could use colorcodes as &l / §l for thick or &o / §o for italic letters also an in-string color changing for example with the layout: "{\0xffffff}Hello {\0x000000}Goodbye"

PR #61, opened 2021-04-24
Added message on linking of AR goggles

Olfi01:


Seniorendi:
Because there are conflicts which I already solved, I will add this feature via a commit.
Thank you anyways!

Issue #77, opened 2021-05-22
[Issue] rsBridge.getItem() doesn't work


Xenxes:
**Describe**

if item doesn't exist in RS storage, getItem returns nil
but if item exists - returns `java.lang.NullPointerException`

**Reproduce**

1. Connect RS Bridge to CC, RS
2. Execute in lua interpreter
```lua
peripheral.find('rsBridge').getItem({ name = 'minecraft:granite' })
```
3. Returns `nil`
4. Add vanilla minecraft "Granite" to RS network
5. Execute same script
6. Returns `java.lang.NullPointerException`

**Versions:**

- Forge version: 36.1.13
- AdvancedPeripherals version: 0.6r
- Refined Storage version: 1.9.12

Seniorendi:
Hey, could you send me your latest log file?

XenProject:
```
[ComputerCraft-Coroutine-0/ERROR]: Error calling call on dan200.computercraft.core.apis.PeripheralAPI@18bda158
java.lang.NullPointerException: null
	at de.srendi.advancedperipherals.common.addons.refinedstorage.RefinedStorage.getItem(RefinedStorage.java:173) ~[?:0.6r]
	at de.srendi.advancedperipherals.common.addons.computercraft.peripheral.RsBridgePeripheral.getItem(RsBridgePeripheral.java:231) ~[?:0.6r]
	at de.srendi.advancedperipherals.common.addons.computercraft.peripheral.RsBridgePeripheral$cc$getItem37.apply(CC generated method) ~[?:?]
	at dan200.computercraft.shared.peripheral.modem.wired.WiredModemPeripheral$RemotePeripheralWrapper.callMethod(WiredModemPeripheral.java:350) ~[?:1.96.0]
	at dan200.computercraft.shared.peripheral.modem.wired.WiredModemPeripheral.callRemote(WiredModemPeripheral.java:177) ~[?:1.96.0]
	at dan200.computercraft.shared.peripheral.modem.wired.WiredModemPeripheral$cc$callRemote29.apply(CC generated method) ~[?:?]
	at dan200.computercraft.core.apis.PeripheralAPI$PeripheralWrapper.call(PeripheralAPI.java:105) ~[?:1.96.0]
	at dan200.computercraft.core.apis.PeripheralAPI.call(PeripheralAPI.java:338) ~[?:1.96.0]
[...]

PR #78, opened 2021-05-23
[#77]Fixed some problems related to the rs bridge.

Xenxes:
I tried to fix it


Seniorendi:
Could you test it if it works?

XenProject:
@Seniorendi, i already did

PR #84, opened 2021-05-26
Fix several issues with AR Goggles

Xenxes:


Seniorendi:
The code looks good, but I'll let Olfi take a look at it.

PR #86, opened 2021-05-30
add DigitalMinerIntegrations

kskudlik:
Adds some integrations for the digital miner (´cause i'm too impatient to wait for mechanism 10.1)


Seniorendi:
Sweet, thank you for the contribution.

PR #88, opened 2021-05-30
fix: InventoryManager removeItemFromPlayer NullPointer if inv is missing

FatalMerlin:
closes Seniorendi/AdvancedPeripherals#87

Just a quick null-check addition.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
@@ -55,7 +55,7 @@ public class InventoryManagerPeripheral extends BasePeripheral {
         IItemHandler inventoryFrom = targetEntity != null ? targetEntity
                 .getCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY, direction).resolve().orElse(null) : null;
         PlayerInventory inventoryTo = getOwnerPlayer().inventory;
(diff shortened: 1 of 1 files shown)


src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
```diff
@@ -55,7 +55,7 @@ public final int addItemToPlayer(String inventoryBlock, int count, Optional<Stri
         IItemHandler inventoryFrom = targetEntity != null ? targetEntity
                 .getCapability(CapabilityItemHandler.ITEM_HANDLER_CAPABILITY, direction).resolve().orElse(null) : null;
         PlayerInventory inventoryTo = getOwnerPlayer().inventory;
-        if (inventoryFrom == null) {
+        if (inventoryFrom == null || inventoryTo == null) {
```
Seniorendi, line 58: You can remove `inventoryFrom == null`. We already check this via `ensurePlayerLinked()`.
And could you remove the debug as well?

Seniorendi (changes requested):

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
FatalMerlin: Done

Seniorendi (approved):

PR #94, opened 2021-06-05
add RadioactiveWasteBarrelIntegration - fix digitalminer reset/start bug

kskudlik:
- add RadioactiveWasteBarrelIntegration
- fix miner reset/start bug (the miner searcher thread couldn't find any ores do to a problem with the way the start method was called)
- changed the signature to be in line with: https://docs.google.com/spreadsheets/d/1sheh1BIeP_vSto881H6UtgXvxdbyJEfILM_uUW_UD7Q/edit#gid=1674588729
 so, that programs will continue to work with Mekanism 10.1 Computer Integration

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/proxyintegrations/mekanism/DigitalMinerIntegration.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/proxyintegrations/mekanism/DigitalMinerIntegration.java
@@ -6,6 +6,11 @@ import mekanism.api.energy.IEnergyContainer;
 import mekanism.api.math.FloatingLong;
 import mekanism.common.tile.machine.TileEntityDigitalMiner;
 
+import mekanism.api.NBTConstants;
+import net.minecraft.nbt.CompoundNBT;
+import mekanism.common.Mekanism;
+import mekanism.common.network.PacketGuiInteract;
+import mekanism.common.network.PacketGuiInteract.GuiInteraction;
 import java.util.List;
 
 public class DigitalMinerIntegration extends ProxyIntegration<TileEntityDigitalMiner> {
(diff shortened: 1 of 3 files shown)


kskudlik:
now this won´t work on dedicated servers...
due to: Attempted to load class net/minecraft/client/Minecraft for invalid dist DEDICATED_SERVER
very unfortunate. so far i´ve no idea why the original way (just calling the start/reset functions) didn´t work.
it´s a bummer

Seniorendi:
Yes, because `sendToServer` from the `packetHandler` from Mekanism calls `SimpleChannel#sendToServer` which calls `Minecraft#getInstance()`. 

But only call `start()` should work, mekanism does that too in V10.1

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/proxyintegrations/mekanism/DigitalMinerIntegration.java
```diff
@@ -26,75 +31,118 @@ protected String getName() {
...
-    public void toggleInverse() {
-        getTileEntity().toggleInverse();
+    public final void stop(){
+        TileEntityDigitalMiner tile = ((TileEntityDigitalMiner) getTileEntity());
+        Mekanism.packetHandler.sendToServer(new PacketGuiInteract(GuiInteraction.STOP_BUTTON, tile));
+    }
+    
+    @LuaFunction
+    public final void reset(){
+        TileEntityDigitalMiner tile = ((TileEntityDigitalMiner) getTileEntity());
```
Seniorendi, line 92: You don't need to create an extra var for the digital miner, you can use `getTileEntity` for `sendToServer`

SirEndii (changes requested):

kskudlik:
in V10.1 it indeed does work. but (as in v10) not with the PeripheralProxy. reset, stop and start does work to some extend but (at least in my testing) when for example resetting with an turtle and starting oresToMine is always 0 and the miner isn´t doing anything.
if i reset and start through the gui it works. the computer can also start and stop as long as the the reset came from the gui.

there is something a little wierd happening cause even in v10.1 calling the start method from the proxy doesn´t work.. hmm

Seniorendi:
I will not close this PR, but will you keep workink on this PR?

SirEndii:
WasteBarrel Integration was added, and the start bug is not fixable. #108

Issue #82, opened 2021-05-25
Writing in chat throws NullPointerException


RedCookieLP:
We set up a modded server with a few mods, including Advanced Peripherals.
Everything worked fine, but for some reason, 2 days prior to today, the chat decided to not work...

I looked at the console of the server and it always throws at least three "NullPointerExceptions"... and it seems to throw them in the "onChatTurtle"-method

I don't know how to reproduce the setup since the problem just popped up out of nowhere and we didn't change anything about the server at the moment the problem first popped up...

- Forge Version: 36.1.13 - MC 1.16.5
- AdvancedPeripherals version: 0.6r

Seniorendi:
Hey, could you send me a full snippet of the errors in the console?

RedCookieLP:
![Exception](https://user-images.githubusercontent.com/45202005/119563325-adeb5000-bda7-11eb-81ee-b020ed1cec57.PNG)

RedCookieLP:
That's the whole exception-block

Seniorendi:
Fixed in 0.6.4b

frank89722:
Still happening in 0.6.4b

```
[29.05 09:32:12] [Server] [Server thread/ERROR] [ne.mi.ev.EventBus/EVENTBUS]: Exception caught during firing event: Cannot invoke "dan200.computercraft.shared.computer.core.ServerComputer.queueEvent(String, Object[])" because the return value of "dan200.computercraft.shared.turtle.blocks.TileTurtle.getServerComputer()" is null
[29.05 09:32:12] [Server] 	Index8
[29.05 09:32:12] [Server] 	Listeners:
[29.05 09:32:12] [Server] 		HIGHEST
[29.05 09:32:12] [Server] 		ASM: class com.feed_the_beast.mods.ftbessentials.FTBEEventHandler playerServerChatHighest(Lnet/minecraftforge/event/ServerChatEvent;)V
[29.05 09:32:12] [Server] 		net.minecraftforge.eventbus.EventBus$$Lambda$2760/0x000000080154b598@5feef21f
[29.05 09:32:12] [Server] 		HIGH
[29.05 09:32:12] [Server] 		ASM: class me.shedaniel.architectury.event.forge.EventHandlerImplCommon event(Lnet/minecraftforge/event/ServerChatEvent;)V
[29.05 09:32:12] [Server] 		NORMAL
[...]

PR #97, opened 2021-06-06
[#82] Fix null pointer on chat events

fxshlein:
It seems like there are cases where CC will return null for `getServerComputer`, my guess is it having to do with turtles somehow being placed in the world without a computer being created (maybe by structures?).

This PR should allow users to chat but the fact that there are tile entities without computers might need further investigation, but I don't really know both mods well enough so I cannot tell if it is needed. CC itself also has null-checks for return values of this function, so it's probably good to have anyways.


Seniorendi:
I already fixed this issue in my dev environment. But your solution is a little bit cleaner
👍🏻 
Thank you for your contribution

fxshlein:
Okay nice!

And no problem, thank you for the mod! 😉

PR #99, opened 2021-06-10
Updated VillagerTrades.java Raised Prices

henkeeman:
I am currently making a custom modpack and I noticed that the computer scientist prices were a bit too low imo. It's completely fine if you liked the old values but I would greatly appreciate a config option to disable the computer scientist profession in that case. Cheers!


Seniorendi:
I will not change the values, but I will create a config for the trades in the future

PR #121, opened 2021-06-29
NBT Storage table writing (and a small API fixes)

SirEdvin:
Add ability to write lua table to nbt storage. 

JSON method are in place, due to json flexibility and more closes nature to NBT than lua table.

Also provides API fixes:
- Variable Store API now mimic to default CC: Tweaked inventory peripheral with slot starting from 1
- For NBT Storage and Geo Scanner configuration can be recived via lua code


Seniorendi:
Thank you for the contribution!

Issue #132, opened 2021-07-03
[Issue] Disabled features on server still show up in game and in creative menu.


Wojbie:
**Describe**

After disabling some features (ME/RS Interface and Chunky Turtle) i expected them to not show up as items and in creative menu? Is there a way to verify they are disabled?

**Reproduce**

Disable feature in server config. Start creative world.

**Screenshots/Videos**

![obraz](https://user-images.githubusercontent.com/5893536/124362302-5e186800-dc34-11eb-8fb2-ee8f231e4a70.png)

**Versions:**

- Forge version:  1.16.5 - 36.1.32
- AdvancedPeripherals version: advancedperipherals-0.6.7b

SirEdvin:
You can try to use them as peripheral) They should just provide no interface, but block presents. 

Personally, I am not sure that idea of removing block after disabling is good, because it can cause player crashes if you disable feature after server works already

SirEdvin:
@Seniorendi May I suggest a solution? Someone can add text note with red color that specific item is disabled. What do you think about it?

Seniorendi:
> @Seniorendi May I suggest a solution? Someone can add text note with red color that specific item is disabled. What do you think about it?

Sounds good

SirEndii:
Fixed in 0.7a

PR #136, opened 2021-07-09
Improvements and fixes

### LINKED ISSUE #20: [Feature] Chat Box: Text components
**Describe your feature idea**
Right now you can only send strings. I think it would be cool if you could send arbitrary components like the `tellraw` command. For example, you could send a clickable component that, when clicked, sends a command for the computer.

Minecraft already has facilities for converting JSON to text components, and back again. Perhaps you could convert a Lua table to JSON, then convert it to a component? And then add some event for receiving incoming text components, which would be like the same but in reverse.

**Describe alternatives you've considered if you've any**
N/A

**Additional context**
`§`-codes can be buggy, and you can't do interactive content like tooltips and click actions.

### LINKED ISSUE #134: [Feature] Environment Sensor
### Describe your idea

The Environment Sensor would allow you to get information about the surrounding entities and blocks. The idea comes from the [Block Scanner](https://plethora.madefor.cc/items/module-scanner.html) and [Entity Sensor](https://plethora.madefor.cc/items/module-sensor.html) from Plethora.
You could get a list of surrounding mobs, for example, and retrieve their meta information (name, position, etc.).

### Describe alternatives you've considered if you've any

_No response_

### Additional context

_No response_

### Linked Issues

_No response_

SirEdvin:
So, this is another pull requests with features! 

- [x] More advancements
- [x] Environmental sensor should have ability to scan entities
- [x] Disable mark for items!
- [x] Turtle display fixes
- [x] Turtle data bridge
- [x] Chat box cooldown returns
- [x] RF feature disabling
- [x] Provide `sendFormattedMessage` to fix #20 

Closes #134, #132, #139


Seniorendi:
It seems that you change the system of our blocks to an energy draining system. But you never drain any energy

SirEdvin:
Hm ... function `consumeFuel` that I call inside scan method, actually, consuming RF from tileEntity with Energy Capability. 

It was done for standardization of energy points, so turtle and blocks can use same logic

Seniorendi:
These changes are impressive. I'll now merge this, cleanup a little bit and commit the last features and fixes for 0.7.

PR #181, opened 2021-08-23
Fixed broken links

HyperCrusher:
All of the metaphysics items in the book recently added seem to point to the items/ subdirectory for api documentation, but the api documentation for these turtles is actually under the metaphysics/ subdirectory


SirEdvin:
Thanks for this merge request!

PR #198, opened 2021-09-09
Add ability to get max item/fluid disk/external storage space.

itzTheMeow:
Added getMaxItemDiskStorage and getMaxFluidDiskStorage as mentioned in SirEndii/Advanced-Peripherals-Features#51.
Both functions return -1 if there is a creative storage disk (as they don't have a max capacity).
Tested using a 1k, 64k, and creative storage disk in one Disk Drive connected to a creative Controller.

Also added getMaxItemExternalStorage and getMaxFluidExternalStorage, not mentioned in the original issue, however requested in the discord server.
Tested using an external storage adapter on a single and double chest connected to a creative controller.

I am not going to implement a function to combine the max disk and external storages as it is redundant, and can be done lua-side.

*also probably good to note this adds for refined storage only, not AE*

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/RsBridgePeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/RsBridgePeripheral.java
@@ -59,6 +59,24 @@ public class RsBridgePeripheral extends BasePeripheral<TileEntityPeripheralOwner
     }
 
     @LuaFunction(mainThread = true)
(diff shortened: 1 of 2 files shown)


SirEdvin:
Hm ... how does it handle filtered external storages and items?

itzTheMeow:
I'm not sure honestly, would be open for someone to test it. I would assume it just returns the total capacity regardless of a filter, because the capacity wouldn't be affected by the filter / it can still hold the same amount of items.

You could fill up a filtered chest with oak planks, its just that only oak planks can enter it.

Issue #177, opened 2021-08-19
Some items do not stack anymore after being removed from the player inventory using the ME bridge


ralphgod3:
### Descripe

when pulling out items from the player inventory using removeItemFromPlayer with the inventory manager some items refuse to stack until placed in the world and broken.

this was tested with both vanilla and moded items.

tested using oak logs, computercraft mining turtles, mekanism logistical transporters.
the logistical transporters stopped stacking, turtles stacked like they normally would but interestingly oak logs also stopped stacking after being pulled out of the inventory.

I did check with the inventory manager list method to see if nbt data changed which it did not
logs dont show anything happening either (should be between 1:20 and 1:30 in the timestamps)

### Steps to reproduce

1 setup inventory manager
2. pull item from player inventory
3. retrieve item from the chest you pushed the item into from player inventory
4. try to get them to stack (this does not happen on all items just some and its not limited to modded items either)

### Multiplayer?

Yes

### Version

0.7r (Latest)

### Minecraft, Forge and maybe other related mods versions

forge 36.2.0 mc 1.16.5 cc 1.98.1

### Screenshots or Videos

https://i.gyazo.com/61e3b969332b28e35db00cd0f74e793d.mp4

### Crashlog/log

https://pastebin.com/8Vb8H7ez

jajajafeti:
I also experienced this bug but with addItemToPlayer I just run the command multiple times, it stacks if u only run it once but I made a Video of the wired behavior.

https://user-images.githubusercontent.com/61298096/132403252-14e4c98f-8a40-4f8f-aa41-18875adaa19e.mp4

jajajafeti:
Hey I just realised this always happens when I use ´getItems()´ with every item in my inventory.
I just wrote a little test programm to test out how ´getItems()´ works and my items never stack after I used it.

https://user-images.githubusercontent.com/61298096/133141207-c3e3ae35-1402-4f5e-8454-70881c01b4d3.mp4

pcfreak9000:
It appears that getItems() adds an nbt-tag to items which usually don't have one. You can see that by enabling the debug tooltips via F3+H and then hovering over the affected items, they will have an empty nbt-tag ("{}") instead of none.

(and i created another duplicate of this issue with possibly more details: #205 )

SirEndii:
Fixed in 0.7.4r

PR #206, opened 2021-10-06
simple fix of issue #177

pcfreak9000:
this should fix issue #177


pcfreak9000:
I don't really understand why the checks fail.

SirEdvin:
Oh, ignore it, this is my fault( I will fix it later, build passed actually (just failed last step with package upload)

PR #209, opened 2021-10-13
Fix: Inventory Manager api, slot argument behaviour is assumedly incorrect.

### LINKED ISSUE #203: Inventory Manager api, slot argument behaviour is assumedly incorrect.
### Description

The Inventory Manager api is dumb/broken.

addItemToPlayer slot designation does not choose where to place an item and instead chooses from which slot to take the item in the originating inventory/chest.
This functionality is stupid because in the alternative function, removeItemFromPlayer, the slot designation chooses the slot in the inventory to take the item from. Thus, I assume the intended behaviour is for addItemToPlayer to also designate a slot in the player inventory to place it.

The api wiki also seems to suggest this is the case, so I assume something went horribly wrong and the behaviour was implemented incorrectly. please fix.

### Steps to reproduce

1) place down computer
2) make lua script to place an item in player inventory via inventory manager.
`manager.addItemToPlayer("UP", 1, 4, "minecraft:dirt")`
3) notice that instead of placing the dirt in slot 4, (in the hotbar according to removeItemFromPlayer), it instead takes the dirt from slot 4 from the chest, and places it in the first empty slot in your inventory.

-- > correct behaviour should be to take from any random slot with the item, or anything if no item designated, and to place said item into slot 4 of the inventory, or return an error if said slot is full/occupied. -1 slot designation should place it in the first empty slot in your inventory.

### Multiplayer?

Yes

### Version

0.7.2r (Latest)

Forge 36.2.6, Minecraft 1.16.5

### Screenshots or Videos

_No response_

### Crashlog/log

_No response_

MrNaru300:
This should fix #203 issue

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
@@ -52,11 +52,6 @@ public class InventoryManagerPeripheral extends BasePeripheral<TileEntityPeriphe
             Item item1 = ItemUtil.getRegistryEntry(item.get(), ForgeRegistries.ITEMS);
             stack = new ItemStack(item1, count);
         }
-        //With this, we can use the item parameter without need to use the slot parameter. If we don't want to use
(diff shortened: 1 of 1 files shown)


jebjordan:
Did not fix.

Issue #237, opened 2021-12-18
Chat box - Formatted message to player, Formatted prefix, User UUIDs, Message Radius


Tim3Game:
### Describe your idea

It's more the just one feature request but they should be mostly easy to implement.
- A `sendFormattedMessageToPlayer(string jsonObject, string user, string prefix)` command, which would allow to send a formatted message to a player only.
- A possibility to have formatted prefixes.
- A possibility to use player's UUID instead of just using their player name (which can changed).
- Ability to set a radius for the `sendMessage()` and `sendFormattedMessage()` functions (so it wouldn't send the message globally).

_If accepted I could try to do them and do a pull request of them if needed, but can't promise anything._

SirEndii:
Added now the following
`sendMessage(String message, int range[, String prefixOrJson, String brackets, String bracketColor])` (Use -1 for infinite range)
Added the same for `ToPlayer` with uuid and name support.

Tim3Game:
So I finally got time to do some code and create a pull request with some changes I made. If you don't like them, fell free to change them.

SirEndii:
Improved in 0.7.7r

PR #240, opened 2021-12-23
Chat box changes

Tim3Game:
Hello,
I saw that you accepted and later added features from the Issue #237. I took the liberty of reviewing it and saw a few things that cough my eye, so I changed them. I wanted to do it sooner myself, but sadly work didn't allowed me to. Hope you don't mind my changes.

**List of changes and additions:**
- replaced `isPlayer(ServerPlayerEntity player, String argument)` with `getPlayer(String argument)` - this allows to get the player (or null if not available) by the username or UUID (if it matches the RegEx for it) instead of looping thru the player list and checking for fitting criterias,
- changes to `appendPrefix(...)` - I changed it from a string to a TextComponent oriented function, this allows for formatting of the prefix itself,
- moved the range argument from the last position to the 3rd one (after the prefix argument), because from my point of view users gonna want to set the range more often the prefix, but that's only my view on it. If it doesn't fit you, you can change it back.
- added "error" messages when the username/uuid is wrong (not found) and when the bracket format is wrong (because from what I saw it expected 2 characters,
- removed the range argument from the "...ToPlayer()" functions, because the range was mostly meant when it sends to multiple players, so it can be set up in a way to send a message to all players in a specific range and not the whole server.

There are other small changes, but they mostly cosmetic code. If you not happy with something, feel free to change it.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
@@ -23,17 +23,19 @@ import net.minecraft.util.Util;
 import net.minecraft.util.text.IFormattableTextComponent;
 import net.minecraft.util.text.ITextComponent;
 import net.minecraft.util.text.StringTextComponent;
+import net.minecraft.util.text.TextComponent;
 import net.minecraftforge.fml.server.ServerLifecycleHooks;
 
 import javax.annotation.Nonnull;
+
+import java.util.Optional;
+import java.util.UUID;
 
 import static de.srendi.advancedperipherals.common.addons.computercraft.operations.SimpleFreeOperation.CHAT_MESSAGE;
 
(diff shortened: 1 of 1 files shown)


SirEndii:
> moved the range argument from the last position to the 3rd one (after the prefix argument), because from my point of view users gonna want to set the range more often the the prefix, but that's only why view on it. If it doesn't fit you, you can change it back.

The reason I didn't do that was to support older programs. Some older programs use the third parameter for the prefix.

Your PR looks good, I will merge it later the day or tomorrow 👍🏻

SirEndii:
Ignore the checks. Build was successful. Only the upload does not work.

Tim3Game:
> The reason I didn't do that was to support older programs. Some older programs use the third parameter for the prefix.

Oh sorry, didn't knew that, because from the wiki the only third parameter was in `sendMessageToPlayer()` and there I removed the `range` argument all together, and in `sendMessage()` and `sendFormattedMessage()` I moved it after the prefix argument to not mess with backwards compatibility.

PR #253, opened 2022-01-07
fix #251

### LINKED ISSUE #251: [BUG] BlockReader gives no usefull data
### Descripe

When I use the BlockRader, it only spits out useless data. I don't think this is intentional.

### Steps to reproduce

1. set up BlockReader with chest at the front.
2. put something into the chest as a test.
3. control the block reader with `getBlockData`

### Multiplayer?

Yes

### Version

0.7.8r (Latest)

39.0.10

### Screenshots or Videos

https://imgur.com/a/KrgM9YR

### Crashlog/log

_No response_

Klebestreifen:
I have fixed my self found bug. I don't know why this is handled so strange in 1.18 but 'save' saves only the metadata and 'saveWithoutMetadata' saves the rest. I tested it ingame with a vanilla BlockEntity (Barrel).

image


SirEndii:
Thank you for the PR
I'll look into this tomorrow. Time for a new 1.18 release.

Issue #263, opened 2022-02-01
AR Clear/Override Specific Elements


ThatGravyBoat:
### Describe your idea

Ability to clear specific elements from the AR Glasses rather than the whole canvas, you can possibly store the canvas as a map of Strings to ARRenderAction instead of a list and then you can maybe do this for example in a script

```lua
local controller = peripheral.find('arController')
local detector = peripheral.find("environmentDetector")

controller.drawItemIcon("minecraft:clock", 4, 4)

while true do
    local timeseconds = math.floor((detector.getTime() % 24000) / 20)
 
    local endTime
    local endTimeNum = math.floor(60*((timeseconds / 60) % 1))
    if endTimeNum < 10 then
        endTime = "0" .. tostring(endTimeNum)
    else
        endTime = tostring(endTimeNum)
    end
 
    controller.clearElement("clock_text")
    controller.drawStringWithId("clock_text", tostring(math.floor(timeseconds / 60)) .. ":" .. endTime, 20, 8, 16777215)
 
    os.sleep(0.25)
end
```

there is one thing about the script above and that is if its a map the drawStringWithId should override the previous one with the same id so that things such as the script I gave in my example would make it so none of it flickers, the clear specific element is so that it doesn't clear the clock when I clear the text but if drawStringWithId overrides the previous one the clear wouldn't be needed in that script but it could be useful for other things.

SirEndii:
Added in 0.7.10b

PR #264, opened 2022-02-05
Added ids to AR Renderer

ThatGravyBoat:
I added ids for render actions as I suggested in #263 

Here are 2 scripts that I tested it with and to show off what it can be used for.
Example 1: This one is a clock that shows up in the top left, it requires both an AR controller and an environment detector
```lua
local controller = peripheral.find('arController')
local detector = peripheral.find("environmentDetector")
 
controller.drawItemIcon("minecraft:clock", 4, 4)
 
while true do
    local timeseconds = math.floor((detector.getTime() % 24000) / 20)
 
    local endTime
    local endTimeNum = math.floor(60*((timeseconds / 60) % 1))
    if endTimeNum < 10 then
        endTime = "0" .. tostring(endTimeNum)
    else
        endTime = tostring(endTimeNum)
    end
 
    controller.drawStringWithId("clock_text", tostring(math.floor(timeseconds / 60)) .. ":" .. endTime, 20, 8, 16777215)
 
    os.sleep(0.25)
end
```
Explanation: if this were done without the drawStringWithId you would need to rerender the clock each time because you would need to do a global clear and that clear would make the text and clock flash.

Example 2: Shows a clock and text that reads "Screen Text" for 5s before removing the text and leaving the clock
```lua
local controller = peripheral.find('arController')
controller.drawItemIcon("minecraft:clock", 4, 4)
controller.drawStringWithId("text", "Screen Text", 20, 8, 0xff00ff)
os.sleep(5)
controller.clearElement("text")
```
Explanation: if this were done without the drawStringWithId again you would need to clear the whole thing and then rerender the clock if you wanted the keep the clock.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ARControllerPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ARControllerPeripheral.java
@@ -50,8 +50,18 @@ public class ARControllerPeripheral extends BasePeripheral<BlockEntityPeripheral
     }
 
     @LuaFunction(mainThread = true)
+    public final void drawStringWithId(String id, String text, int x, int y, int color) {
+        owner.tileEntity.addToCanvas(new ARRenderAction(id, RenderActionType.DrawString, text, x, y, color));
+    }
+
+    @LuaFunction(mainThread = true)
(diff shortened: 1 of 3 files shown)


SirEndii:
Thank you for the PR.
Do you want to change the documentation? If not, I would do it in the next days after I implemented it

ThatGravyBoat:
yeah I can update the documentation too

PR #292, opened 2022-04-26
Fixed ME craft NullPointerException if no item in stock.

### LINKED ISSUE #291: NullPointerException crash if trying to craft an item with 0 stock with ME bridge
### Descripe

Own compiled build from current master as of writing, as AE Bridge is broken in latest release.

Game/server crashes if trying to craft an item is not in stock in the ME system.

I think the bug is here:
https://github.com/Seniorendi/AdvancedPeripherals/blob/8debb40a941977609bd80f9e9baf0b1f12a2f7d8/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/CraftJob.java#L83

aeItem can be null, and .getRight is called just after. :)

### Steps to reproduce

Have a recipe in ME system, but no items in stock/in system. 

```lua
me = peripheral.wrap("left")
me.craftItem({name="minecraft:oak_planks", count=10})
```

### Multiplayer?

No

### Version

Own compiled build.

Forge 40.1.0, Minecraft 1.18.2, AE2 11.0.0-alpha.6

### Screenshots or Videos

_No response_

### Crashlog/log

https://pastebin.com/298NPaBF

### LINKED ISSUE #293: [ME Bridge] isItemCraftable, listCraftableItems, isItemCrafting bug if no item stack of items in ME system
### Descripe

`isItemCraftable`

- Does not return true, if an item is craftable, but the amount in the ME system is 0.

`listCraftableItems`

- Does not return items that are indeed craftable, but the amount is 0.

`isItemCrafting`

- Gives lua error (Java null pointer exception) if amount is 0.

All these errors are caused by the same bug as #291.

I already have a fix local, but it is blocked waiting on the method I added in #292 :) Diff can be seen: https://github.com/LauJosefsen/AdvancedPeripherals/compare/fix-null-pointer-exception-me-craft-no-stock...LauJosefsen:fix-items-not-being-craftable-if-0-stock

### Steps to reproduce

Have an item craftable but with amount = 0.

Call isItemCraftable, listCraftableItems or isItemCrafting on the item.

### Multiplayer?

No

### Version

Own compiled build.

Forge 40.1.0, Minecraft 1.18.2, AE2 11.0.0-alpha.6

### Screenshots or Videos

_No response_

### Crashlog/log

_No response_

LauJosefsen:
Refactored the method to early returns instead of break.

Added overload with crafting service, to search for craftable items to find AEItemKeys.

Closes #291

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
@@ -29,18 +29,34 @@ import java.util.Map;
 
 public class AppEngApi {
 
-    public static Pair<Long, AEItemKey> findAEStackFromItemStack(MEStorage monitor, ItemStack item) {
-        Pair<Long, AEItemKey> stack = null;
+    public static Pair<Long, AEItemKey> findAEStackFromItemStack(MEStorage monitor, ICraftingService crafting, ItemStack item) {
         for (Object2LongMap.Entry<AEKey> temp : monitor.getAvailableStacks()) {
             if (temp.getKey() instanceof AEItemKey key) {
                 if (key.matches(item)) {
-                    stack = Pair.of(temp.getLongValue(), key);
-                    break;
+                    return Pair.of(temp.getLongValue(), key);
                 }
             }
         }
-        return stack;
+
+        if (crafting == null)
+            return null;
+
+        for (var temp : crafting.getCraftables(param -> true)) {
+            if(temp instanceof AEItemKey key) {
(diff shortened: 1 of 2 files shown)


SirEndii:
Looks good to me!
I'll merge this when I get the time

LauJosefsen:
Awesome!

Also, I found that isCraftable and listCraftableItems also does not work if item is not in stock, but is craftable.

I have made a branch that fixes that as well, but it is branched out from this branch. You can see diff here if interested: https://github.com/LauJosefsen/AdvancedPeripherals/pull/1/files

SirEndii:
Thank you for your contribution!

PR #295, opened 2022-04-27
Geo Scanner height, fixes issue #266

### LINKED ISSUE #266: Geo Scanner chunkAnalyze() function does not include negative y in range
### Descripe

Looking at the code for the Geo Scanner peripheral here: https://github.com/Seniorendi/AdvancedPeripherals/blob/a920c50f6b4a1f210ab27b951298168d70325fd0/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/GeoScannerPeripheral.java

The chunkAnalyze() function checks between y values of 0 and 256. In Minecraft 1.18.1, y values go below 0. I think it would be best to include these negative values in this function as many important ores are located there.

### Steps to reproduce

Look at line 111 of code here: https://github.com/Seniorendi/AdvancedPeripherals/blob/a920c50f6b4a1f210ab27b951298168d70325fd0/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/GeoScannerPeripheral.java

### Multiplayer?

Yes

### Version

1.18.1-0.7.9r (Latest 1.18)

Forge 39.0.59 Minecraft 1.18.1

### Screenshots or Videos

_No response_

### Crashlog/log

_No response_

JensenJ:
Fixed an issue where the geo scanner would not make use of 1.18 terrain heights in the overworld. This should work for every dimension regardless of its height. 

This should also fix #266.


SirEndii:
Thank you for your contribution! ^^

PR #296, opened 2022-04-28
Is item crafting fix by iterating cpus

### LINKED ISSUE #293: [ME Bridge] isItemCraftable, listCraftableItems, isItemCrafting bug if no item stack of items in ME system

### LINKED ISSUE #294: [ME Bridge] isItemCrafting not working as hinted by documentation
### Descripe

Hi again!

I have found that the method `isItemCrafting` probably does not do what most users would expect. I can see internally the AE2 api `isRequested` is used, however that API seems to only return true if the item is currently crafting, and not "scheduled"

What I'm trying to say is, if you craft, for example, a stone button. The ME system might need to smelt some stone first. While the stone is smelting, `isRequested` for stone button will return _false_. Reading the documentation of your mod, I don't think this is intended behavior, as a job for stone button does in fact request.

If wanting to fix, I propose iterating through crafting CPUs to see what they are crafting. Something like this: https://github.com/LauJosefsen/AdvancedPeripherals/compare/fix-items-not-being-craftable-if-0-stock...LauJosefsen:isItemCrafting_fix_by_iterating_cpus

Hope I don't annoy you with these issues 😄 

### Steps to reproduce

1.  Setup a crafting pattern for stone button in ME system
2.  Setup a smelting pattern for cobblestone-stone in ME system
3.  While the system is waiting for stone (and the status in the me system is "scheduled"), `isItemCrafting({name="minecraft:stone_button"}) will return false.

### Multiplayer?

No

### Version

Own compiled build.

Forge 40.1.0, Minecraft 1.18.2, AE2 11.0.0-alpha.6

### Screenshots or Videos

_No response_

### Crashlog/log

_No response_

LauJosefsen:
Closes #294  by iterating over crafting cpus instead of using `isRequested` api.

Closes #293
- I can see you closed #293 yesterday, but it was not fixed by #292, sorry for the confusion. 😅

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
@@ -1,6 +1,7 @@
 package de.srendi.advancedperipherals.common.addons.appliedenergistics;
 
 import appeng.api.networking.IGridNode;
+import appeng.api.networking.crafting.CraftingJobStatus;
 import appeng.api.networking.crafting.ICraftingCPU;
 import appeng.api.networking.crafting.ICraftingPlan;
 import appeng.api.networking.crafting.ICraftingService;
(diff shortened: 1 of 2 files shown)


SirEndii:
👍🏻

PR #318, opened 2022-07-21
Added inventory manager nbt support

### LINKED ISSUE #317: InventoryManager add/remove with NBT data
### Describe your idea

Currently the InventoryManager's `addItemToPlayer` and `removeItemFromPlayer` function's item argument does not support NBT tags, thus we cannot add/remove enchanted items or rockets for example.  
It would be nice to:
- Add new functions, as to not change the existing API, or
- Change the functions' signature to instead take a table, like the RS/ME bridges' `exportItem`

### Describe alternatives you've considered if you've any

One alternative is to export from the  RS/ME system to an empty chest, and don't filter while using the InventoryManager

### Additional context

I think the second option is better even tho it changes API, because it would unify the ME and RS bridges and the InventoryManager

### Linked Issues

_No response_

Erv00:
Added two functions to work with nbt, closes #317 
The old functions are unchanged, the new ones work the same, but support nbt tags.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/InventoryManagerPeripheral.java
@@ -2,6 +2,8 @@ package de.srendi.advancedperipherals.common.addons.computercraft.peripheral;
 
 import dan200.computercraft.api.lua.LuaException;
 import dan200.computercraft.api.lua.LuaFunction;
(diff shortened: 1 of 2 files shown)


SirEndii:
I'll review this when I get back home in a week

SirEndii:
Looking good - I don't see any issues so far

SirEndii:
Would you want to add this to the ap documentation? https://github.com/Seniorendi/Advanced-Peripherals-Documentation

Erv00:
Yes, I'll do it soon

PR #332, opened 2022-08-21
Use item data from each slot used to preserve NBT data.

aowolfie:
A small change to preserve item NBT data when importing from a target inventory using a ME bridge.


aowolfie:
Note that this bug may affect other storage devices like the RS interface.

Issue #362, opened 2022-11-04
Me Bridge's listCraftableItems() does not behave as expected.


toastonrye:
### Describe

minecraft-forge 1.19.2-43.1.47
cc-tweaked-1.19.2-1.101.1.jar
appliedenergistics2-forge-12.8.4.jar

AdvancedPeripherals-0.7.21b.jar

listCraftableItems() only returns a table if the item amount is greater than 0. If there is 0 of some item that is craftable, it should still return a table, with amount 0.

### Steps to reproduce

```
local me = peripheral.find("meBridge")
if not me then error("meBridge not found") end

patterns = me.listCraftableItems()
for k, v in pairs(patterns) do
 print("listCraftableItems: ", v)
end
for k, items in pairs(patterns) do
 print(items.name, items.amount)
end
```

53 second demonstration
https://www.youtube.com/watch?v=iQxfN4farMA

### Multiplayer?

No

### Version

1.19.2-0.7.19a (Latest 1.19)

### Minecraft, Forge and maybe other related mods versions

minecraft-forge 1.19.2-43.1.47

### Screenshots or Videos

https://www.youtube.com/watch?v=iQxfN4farMA

SirEndii:
Reopen until released
This issue is now fixed for 1.19 and 1.18

I sadly can't build you a test file since the current branch is on 1.19.3, and we're unable to build for this version until some dependencies update.

PR #377, opened 2022-12-07
Fix for Issue #362

UnhappyCodings:
Added only craftable and not in AE2 ME exisiting items to listCraftableItems() list with count of 0.


src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
```diff
@@ -60,8 +63,16 @@ public static Pair<Long, AEItemKey> findAEStackFromItemStack(MEStorage monitor,
 
     public static List<Object> listStacks(MEStorage monitor, ICraftingService service, int flag) {
         List<Object> items = new ArrayList<>();
-        for (Object2LongMap.Entry<AEKey> aeKey : monitor.getAvailableStacks()) {
+        KeyCounter keyCounter = monitor.getAvailableStacks();
+        for (Object2LongMap.Entry<AEKey> aeKey : keyCounter) {
             if (aeKey.getKey() instanceof AEItemKey itemKey) {
+                if (flag == 2)
```
SirEndii: Please use the if statement we already use on line 78

PR #404, opened 2023-01-25
Added some methods to the MEBridge

michele-grifa:
I have added the following method to the bridge.

- **getTotalItemStorage**: Return the total number of bytes for items, installed in the system;

- **getTotalFluidStorage**: Return the total number of bytes for fluids, installed in the system;

- **getUsedItemStorage**: Return the used number of bytes for items;

- **getUsedFluidStorage**: Return the used number of bytes for fluids;

- **getAvailableItemStorage**: Return the total number of bytes, available for storage;

- **getTotalFluidStorage**: Return the total number of bytes for fluids, available for storage;

- **listCells**: The method return a table with the basic info of all cells connected to the system. The basic info are:
  - item: item name
  - cellType: fluid or item
  - bytesPerType
  - totalBytes

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
@@ -1,5 +1,6 @@
 package de.srendi.advancedperipherals.common.addons.appliedenergistics;
 
+import appeng.api.inventories.InternalInventory;
 import appeng.api.networking.IGridNode;
 import appeng.api.networking.crafting.CraftingJobStatus;
 import appeng.api.networking.crafting.ICraftingCPU;
@@ -11,7 +12,10 @@ import appeng.api.stacks.AEItemKey;
 import appeng.api.stacks.AEKey;
 import appeng.api.stacks.KeyCounter;
 import appeng.api.storage.AEKeyFilter;
+import appeng.api.storage.IStorageProvider;
 import appeng.api.storage.MEStorage;
+import appeng.blockentity.storage.DriveBlockEntity;
+import appeng.items.storage.BasicStorageCell;
 import dan200.computercraft.shared.util.NBTUtil;
 import de.srendi.advancedperipherals.AdvancedPeripherals;
 import de.srendi.advancedperipherals.common.util.LuaConverter;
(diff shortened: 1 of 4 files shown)


SirEndii:
Thank you for your contribution!
Ignore the failed checks, something is wrong with my keys.
I will check that tomorrow. 🙂

michele-grifa:
👍

michele-grifa:
I tried to use the mod with AE2 Things Deep Item Storage Disk and i noticed that i received wrong values, so i added support for it.

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
```diff
@@ -228,4 +229,205 @@ public static boolean isItemCrafting(MEStorage monitor, ICraftingService grid, I
...
+            InternalInventory inventory = entity.getInternalInventory();
+
+            for(int i = 0; i < inventory.size(); i++) {
+                ItemStack stack = inventory.getStackInSlot(i);
+
+                if(stack.getItem() instanceof BasicStorageCell cell) {
+                    if(cell.getKeyType().toString().equals("ae2:i")) {
+                        total += cell.getBytes(null);
+                    }
+                } else if(stack.getItem() instanceof DISKDrive disk) {
```
SirEndii: A minecraft instance without this mod installed would crash as soon I call this method.
I would recommend to check if the stack is empty, and if it is, continue to the next item.
This would prevent that "instanceof" checks that the "air" item in an empty item stacks is a disk drive


Same for the `listCells` method

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
```diff
@@ -228,4 +229,205 @@ public static boolean isItemCrafting(MEStorage monitor, ICraftingService grid, I
...
+            if(entity == null) continue;
+
+            InternalInventory inventory = entity.getInternalInventory();
+
+            for(int i = 0; i < inventory.size(); i++) {
+                ItemStack stack = inventory.getStackInSlot(i);
+
+                if(stack.getItem() instanceof BasicStorageCell cell) {
+                    items.add(index, getObjectFromCell(cell, stack));
+                    index++;
```
SirEndii: Why exactly do you use an index here instead of just adding it to the list?

SirEndii (changes requested):

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
michele-grifa: I added the needed checks.

SirEndii (approved):

SirEndii:
Looking good!

PR #405, opened 2023-01-26
Added methods to import/export fluid from/to the ME System

michele-grifa:
New methods added:

- `importFluid(table fluid, string direction)`
- `importItemFromPeripheral(table item, string tank)`
- `exportFluid(table fluid, string direction)`
- `exportFluidFromPeripheral(table item, string tank)`

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
@@ -90,7 +90,7 @@ public abstract class PeripheralBlockEntity<T extends BasePeripheral<?>> extends
 
         if (cap == CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY && !remove && direction != null) {
             if (fluidHandler == null || !fluidHandler.isPresent())
-                fluidHandler = LazyOptional.of(() -> new FluidTank(1000));
+                fluidHandler = LazyOptional.of(() -> new FluidTank(0));
             return fluidHandler.cast();
         }
         return super.getCapability(cap, direction);
--- a/src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
@@ -23,6 +23,9 @@ import net.minecraft.world.level.block.entity.BlockEntityType;
 import net.minecraft.world.level.block.state.BlockState;
 import net.minecraftforge.common.capabilities.Capability;
 import net.minecraftforge.common.util.LazyOptional;
+import net.minecraftforge.fluids.FluidUtil;
+import net.minecraftforge.fluids.capability.CapabilityFluidHandler;
+import net.minecraftforge.fluids.capability.IFluidHandler;
 import net.minecraftforge.items.CapabilityItemHandler;
 import net.minecraftforge.items.IItemHandler;
 import net.minecraftforge.items.wrapper.SidedInvWrapper;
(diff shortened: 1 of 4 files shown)


src/main/java/de/srendi/advancedperipherals/common/util/FluidUtil.java
```diff
@@ -0,0 +1,147 @@
...
+import javax.annotation.Nullable;
+import java.util.ArrayList;
+import java.util.List;
+import java.util.Map;
+import java.util.Objects;
+import java.util.function.Function;
+import java.util.function.Predicate;
+
+public class FluidUtil {
+    private static final List<Pair<Predicate<Object>, Function<Object, IFluidHandler>>> EXTRACTORS = new ArrayList<>();
```
SirEndii: For what have you added this? Is there a specific reason? If yes, a comment should be here

michele-grifa: There is no specific reason, I copied from the InventoryUtil class and I forgot to delete this piece. I removed it.

SirEndii: Alright

src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
```diff
@@ -82,6 +86,12 @@ public <T1> LazyOptional<T1> getCapability(@NotNull Capability<T1> cap, @Nullabl
                 handler = LazyOptional.of(() -> new SidedInvWrapper(this, Direction.NORTH));
             return handler.cast();
         }
+
+        if (cap == CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY && !remove && direction != null) {
+            if (fluidHandler == null || !fluidHandler.isPresent())
+                fluidHandler = FluidUtil.getFluidHandler(this.getLevel(), this.getBlockPos(), Direction.SOUTH);
```
SirEndii: This leads to a StackOverflowError since `getFluidHandler` calls `getCapability` of the tile entity

SirEndii (changes requested):

src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
SirEndii: You might want to create a new fluid handler or want to use `net.minecraftforge.fluids.capability.TileFluidHandler`

SirEndii (approved):

src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
michele-grifa: I created a new FluidTank. If I can ask, how I can implements with TileFluidHandler? I should replace 
`private LazyOptional<? extends IFluidHandler> fluidHandler;` with `private LazyOptional<? extends TileFluidHandler> fluidHandler;` ?

SirEndii: The way with the fluid tank is simpler and more than enough, I just forgot that the fluid tank implements a fluid handler

michele-grifa: Ok thanks.

src/main/java/de/srendi/advancedperipherals/common/blocks/base/PeripheralBlockEntity.java
```diff
@@ -82,6 +87,12 @@ public <T1> LazyOptional<T1> getCapability(@NotNull Capability<T1> cap, @Nullabl
                 handler = LazyOptional.of(() -> new SidedInvWrapper(this, Direction.NORTH));
             return handler.cast();
         }
+
+        if (cap == CapabilityFluidHandler.FLUID_HANDLER_CAPABILITY && !remove && direction != null) {
+            if (fluidHandler == null || !fluidHandler.isPresent())
+                fluidHandler = LazyOptional.of(() -> new FluidTank(1000));
```
SirEndii: Everything seems to work fine. But the capacity should be zero, since the bridge does not do anything with the contained fluid. The bridge would just store one bucket of liquid forever.

SirEndii (changes requested):
Should be the last thing

SirEndii (approved):

PR #462, opened 2023-05-18
Add configurable chunk loading radius to chunky turtle

Einhornyordle:
I've added a configurable chunk loading radius for the chunky turtle as requested in SirEndii/Advanced-Peripherals-Features#39.


src/main/java/de/srendi/advancedperipherals/common/configuration/PeripheralsConfig.java
```diff
@@ -81,6 +83,7 @@ public PeripheralsConfig() {
 
         enableChunkyTurtle = builder.comment("Enable the Chunky Turtle or not.").define("enableChunkyTurtle", true);
         chunkLoadValidTime = builder.comment("Time in seconds, while loaded chunk can be consider as valid without touch").defineInRange("chunkLoadValidTime", 600, 60, Integer.MAX_VALUE);
+        chunkyTurtleRadius = builder.comment("Radius in chunks a single chunky turtle will load. The default value (0) only loads the chunk the turtle is in, 1 would also load the 8 surrounding chunks (9 in total) and so on").defineInRange("chunkyTurtleRadius", 0, 0, Integer.MAX_VALUE);
```
SirEndii: `Integer.MAX_VALUE` would be a bad max value for the config value
Something like 16 sounds more reasonable to me

SirEndii (changes requested):

src/main/java/de/srendi/advancedperipherals/common/configuration/PeripheralsConfig.java
Einhornyordle: Done

SirEndii (approved):

SirEndii:
@Patprehn poke

PR #467, opened 2023-05-30
Added playerJoin, playerLeave, playerChangedDimension events on Playe…

michele-grifa:
…r Detector

Based on SirEndii/Advanced-Peripherals-Features#10

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/PlayerDetectorPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/PlayerDetectorPeripheral.java
@@ -17,8 +17,6 @@ import de.srendi.advancedperipherals.lib.peripherals.BasePeripheral;
 import net.minecraft.core.BlockPos;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.entity.player.Player;
-import net.minecraftforge.common.MinecraftForge;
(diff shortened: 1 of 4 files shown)


SirEdvin:
I just want to add, that this effectively register every peripheral and don't de-register them on detach. I am not sure if Forge will handle it correctly ... 

Probably it is better to use approach from [Events](https://github.com/SirEndii/AdvancedPeripherals/blob/1.19.3/src/main/java/de/srendi/advancedperipherals/common/events/Events.java)

michele-grifa:
Yes but if i use that approach, the methods must be static and so I don't have access to the current peripheral.

SirEdvin:
Yes, and in [ChatBox](https://github.com/SirEndii/AdvancedPeripherals/blob/1.19.3/src/main/java/de/srendi/advancedperipherals/common/blocks/blockentities/ChatBoxEntity.java) you subscribe to this message queue and read it every tick. 
And from block entity you have access to peripheral.

michele-grifa:
I used the Events approach, now should be better.

The example for the docs are:
Player Join Event
```lua
local event, playerName, dimension = os.pullEvent('playerJoin')
```

Player Leave Event
```lua
local event, playerName, dimension = os.pullEvent('playerLeave')
```

Player Changed Dimension Event
```lua
local event, playerName, fromDim, toDim = os.pullEvent('playerChangedDimension')
```

SirEndii:
Thank you!

PR #488, opened 2023-07-14
Introduce 1.106 support

SirEdvin:
So, this is basically 1.106 support. Automata cores are also stateful, so initial cooldown logic probably can be removed.


SirEndii:
Thanks edvin

SirEndii (approved):

SirEdvin:
You're welcome

PR #540, opened 2023-12-21
Add peripheral functions for the Dim Chest of the Dimensional Storage…

Einhornyordle:
… mod. Also add Jade for better testing purposes.

The GUI configuration options of the dimensional chest from the dimensional storage mod are now exposed as a peripheral, which allows you to configure the chest as a peripheral.

The chest is exposed as a default inventory peripheral, allowing only to access it's inventory.

The chest now exposes its configuration instead of the inventory.

Technically, since now you can't access the inventory any more.

* **Other information**:
As we talked before, you still need to implement a way to expose the inventory and the configuration at the same time, since multiple peripheral exposures are not possible at the moment.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -306,6 +306,12 @@ dependencies {
     implementation fg.deobf("me.shedaniel.cloth:cloth-config-forge:11.1.106")
     implementation fg.deobf("dev.architectury:architectury-forge:9.1.10")
 
+    //DimStorage
+    compileOnly fg.deobf("curse.maven:dimstorage-353882:${dimstorage_version}")
+    runtimeOnly fg.deobf("curse.maven:dimstorage-353882:${dimstorage_version}")
+    compileOnly fg.deobf("curse.maven:edivadlib-638508:${edivadlib_version}")
+    runtimeOnly fg.deobf("curse.maven:edivadlib-638508:${edivadlib_version}")
+
     //Removed until fully ported
     //testImplementation fg.deobf("site.siredvin.ttoolkit:ttoolkit-${minecraft_version}:${ttoolkit_version}")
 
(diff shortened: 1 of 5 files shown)


build.gradle
```diff
@@ -320,6 +326,8 @@ dependencies {
     // Testing stuff
     // JEI
     implementation fg.deobf("mezz.jei:jei-${jei_version}")
+    // Jade
+    implementation fg.deobf("curse.maven:jade-324717:${jade_version}")
```
SirEndii, line 330: Please don't use cursemaven for jade

src/main/java/de/srendi/advancedperipherals/common/addons/dimstorage/DimChestIntegration.java
```diff
@@ -0,0 +1,55 @@
...
+        super(entity);
+    }
+
+    @NotNull
+    @Override
+    public String getType() {
+        return "dimChest";
+    }
+
+    @LuaFunction
```
SirEndii: Please set `mainThread` to true to be thread safe with the world
`@LuaFunction(mainThread = true)`

SirEndii (changes requested):

build.gradle
Einhornyordle: There is no other option

src/main/java/de/srendi/advancedperipherals/common/addons/dimstorage/DimChestIntegration.java
Einhornyordle: Fixed

build.gradle
SirEndii: Cursed

SirEndii (approved):

PR #549, opened 2024-01-13
Add entity API

lonevox:

Feature.

Calling `EnvironmentDetectorPeripheral.scanEntities()` gives only a small amount of information on each entity. I felt like this could be used to a greater potential.

Instead of getting a table with a small amount of data when calling `EnvironmentDetectorPeripheral.scanEntities()`, UUIDs for entities are now the only thing that is returned. These can be used with the new entity API. For example, `entity.getNBT(uuid)` returns the NBT data of the entity with the given UUID.
The entity API itself can be disabled in config, as well as the individual functions within it that could be deemed exploitative on servers.
A `PlayerDetectorPeripheral.getPlayerUUID(username)` Lua function has also been added so that you can make use of the entity API when using a `PlayerDetectorPeripheral`.
I have also expanded the amount of entity information you get when calling `LuaConverter.completeEntityToLua(entity)`. This function can be called via the entity API with `entity.getData(uuid)`.

Yes, any calls to `EnvironmentDetectorPeripheral.scanEntities()` will need to use the entity API to get the data that it used to give you.
Calls to `AutomataEntityHandPlugin.searchAnimals()` now have position information in a table with the key `relativePos`. This was a side-effect of cleaning up `LuaConverter.completeEntityToLua(entity)` to not require an `ItemStack` to be passed to it.

* **Other information**:
I can make a PR for the docs if this PR is to be merged.


lonevox:
Here's a script that I've been testing with: https://pastebin.com/uCH39VEm

SirEndii:
Before I start reviewing

Since this adds a wider palette of features and due to the fact that this includes a breaking change, could you change the target branch to `dev/0.8` so this features will be implemented to the next major version?

lonevox:
Done

SirEndii:
Greetings

The build seems to fail, I am not sure if the issues came with the change of the target branch, but they are there
Could you fix them? I would also recommend running gradle check afterward to check for checkstyle rule violations.

You can either open the build log from teamcity by clicking on "Details" or you just could try to build it locally

lonevox:
My bad, it builds now

SirEndii:
You still didn't fix the [checkstyle violations](https://teamcity.intelligence-modding.de/buildConfiguration/Advancedperipherals_BuildAndPublish/5139?hideTestsFromDependencies=false&hideProblemsFromDependencies=false&expandBuildDeploymentsSection=false&expandBuildChangesSection=true&expandBuildProblemsSection=true&expandPull+Request+Details=true)
Either check the teamcity build log or run `gradle check` locally

SirEndii:
Foremost, thanks for the PR and I really like your idea
I also discussed that internally with some collaborators, and we agree that switching to an LuaAPI based entity api would be a nice addition to AP

The changes look good, but I didn't inspect them too deeply. I will take some time this week to review your changes and test them a bit for me

tehgreatdoge:
Would it have made more sense to implement this with peripheral methods instead of polluting the global namespace? You could return a userdata from the scan method, return all the information by default, or just add another method to the peripheral for looking up data by uuid

lonevox:
> Would it have made more sense to implement this with peripheral methods instead of polluting the global namespace?

Originally I started with adding a `getNBT(UUID)` function to `EnvironmentDetectorPeripheral`, since I didnt want to put the huge NBT information in the small table returned by `scanEntities`, but if you think about it it makes no sense to have a getNBT function in `EnvironmentDetectorPeripheral`. It was either have a massive table returned or separate this into an Entity API which can have more fine-grained functions, so I opted for the later. It also isn't clear which functions in different peripherals return what entity data, but if you get a UUID from `EnvironmentDetectorPeripheral.scanEntities` and a UUID from `PlayerDetectorPeripheral.getPlayerUUID`, you know that `Entity.getData(UUID)` will return the same type of information for both entities.

> You could return a userdata from the scan method

It was already the case before that some functions returned different entity data than others for no real reason. `scanEntities` only gave you like 10 fields per entity so it wasn't very useful, whereas `searchAnimals` gave you a lot of information, but still didn't include NBT data. It is easier to maintain and more consistent if this is all in one place.

> return all the information by default

You could return all data, but there is a *lot* of information available now. It feels almost wrong to lump almost all information on the entity into one huge table. This would be NBT, almost all runtime data, and all persistent data. If you just need to get the positions of entities then this is all useless information.

> or just add another method to the peripheral for looking up data by uuid

Something like `getEntityData(UUID)` wouldn't make sense in any one peripheral

lonevox:
After looking at some of my changes again though, I think it would make sense for `AutomataEntityHandPlugin.searchAnimals` and `AutomataEntityTransferPlugin.getCapturedAnimal` to both return UUIDs instead of the result of `completeEntityToLuaWithShearable` to be consistent with my other changes

SirEndii:
> Would it have made more sense to implement this with peripheral methods instead of polluting the global namespace? You could return a userdata from the scan method, return all the information by default, or just add another method to the peripheral for looking up data by uuid

I am currently discussing that a bit with some of the CC community in the CC discord. I know that additions to the global API are not welcome in the community. AP already does unconventional things that I want to eliminate in 0.8, and I'd hate to add more.

lonevox:
> I am currently discussing that a bit with some of the CC community in the CC discord. I know that additions to the global API are not welcome in the community. AP already does unconventional things that I want to eliminate in 0.8, and I'd hate to add more.

Damn thats unfortunate, I wasn't aware of that. It's an issue of usability and code quality vs global pollution then. It's a shame that an entity API isn't available in CC itself.

I guess a rough compromise could be standardizing the entity data returned in different parts of this mod. So pretty much chucking everything into one huge table, but making sure that happens consistently in every peripheral function where an entity is retrieved. It would be worth documenting this standard entity table somewhere.

Another alternative is turning the Player Detector peripheral into an Entity Detector peripheral. All functionality of the entity API could be put in the Entity Detector. `scanEntities` could be removed from the Environment Detector and put into the Entity Detector.

SirEndii:
> Another alternative is turning the Player Detector peripheral into an Entity Detector peripheral. All functionality of the entity API could be put in the Entity Detector. scanEntities could be removed from the Environment Detector and put into the Entity Detector.

I support that idea
Adding all the entity related features to one block would be the ideal compromise imo

you could also participate in the small discussion in the #modmaking chat on the [cc discord](https://discord.computercraft.cc/)

SirEndii:
One idea I like is to still add all the entity API functions to the environmental detector but to separate them a bit, put them into a table which you can retrieve using `getEntityAPI`.
This would be simpler than "adding the API to the global environment, but only when the peripheral is attached" imo

So maybe you want to go with that?

I will also convert this to a draft until we agree on a method

lonevox:
> One idea I like is to still add all the entity API functions to the environmental detector but to separate them a bit, put them into a table which you can retrieve using `getEntityAPI`. This would be simpler than "adding the API to the global environment, but only when the peripheral is attached" imo
> 
> So maybe you want to go with that?
> 
> I will also convert this to a draft until we agree on a method

For the idea of turning the Player Detector into an Entity Detector:
Player Detector has four player-specific things: the `playerClick`, `playerJoin`, `playerLeave` events, and the `getOnlinePlayers` function. Everything else can be made to work for any entity (1 event and 10 functions). The `getPlayers...` functions currently return the names of players, but they could become `getEntities...` functions and return entity UUIDs instead. The functions migrated from the `EntityAPI` like `getName(UUID)` or `getNBT(UUID)` would have checks within them to see if the entity with the UUID is still within the range of the Entity Detector so that these can't be abused to get information about entities anywhere.

I think a `getEntityAPI` on the Environment Detector is a possibility, although you'd then have almost duplicate functions across peripherals (such as `EnvironmentDetector.EntityAPI.getPos` and `PlayerDetector.getPlayerPos`). An Entity Detector could replace all functions that get entity data in all other peripherals.

The biggest consideration with the potential Entity Detector is if its okay for it to have the 4 player-specific things. These 4 things cant really be removed, so its just if it makes sense or not. If it doesn't make sense, then it might not make sense to turn the Player Detector into an Entity Detector. I'm not sure, I'd like your opinion.

SirEndii:
Merging the two peripherals is an interesting idea, but I don't really like it. Merging them would create one peripheral with maybe too many functions imo which wouldn't keep it simple

Developing the Player Detector and the Environment Detector separately would be my way to go. I would simply accept the duplicates, as there are apparently only 2 functions (getPlayerPos and getOnlinePlayers)

Keeping the player related functions (isPlayerInX... getPlayersInX...) and events to the player detector would make it simpler for the user imo to just have player related functions in one dedicated block while there would be the more advanced entity api in the Environment Detector

lonevox:
Sweet. I'm going away for a week but I'll work on that when I get back

SirEndii:
Greetings @lonevox 
Do you have any plans on continuing this?

lonevox:
> Greetings @lonevox Do you have any plans on continuing this?

I took quite a hiatus from making my pack which is why I completely disappeared, sorry about that. I'll try getting the PR ready with the discussed changes within the next few days.

lonevox:
@SirEndii I'm having trouble exposing the EntityAPI class as a table through the EnvironmentDetectorPeripheral like you suggested. Here's what I assume would work, but isn't:
```java
@LuaFunction(mainThread = true)
public final EntityAPI getEntityAPI() {
    return new EntityAPI();
}
```
It successfully returns a table, but the table is empty, even though there are 6 LuaFunctions in EntityAPI. Any ideas?

SirEndii:
Can you check the log?
CC throws some debug messages when it's unable to parse functions

SirEndii:
I currently do something similar for the smart glasses
Everything should work fine if you don't have invalid arguments or return types in your function

https://github.com/IntelligenceModding/AdvancedPeripherals/blob/637e1749a5e89c3287a7d6dd0eb2c38fa1178224/src/main/java/de/srendi/advancedperipherals/common/smartglasses/modules/nightvision/NightVisionFunctions.java#L6-L22

lonevox:
I'm pretty sure the function arguments and return types are fine, especially as I was using these methods fine when using EntityAPI as a CC API.

Even this for the EntityAPI isn't working:
```java
public class EntityAPI {
    @LuaFunction
    public final String test() {
        return "Hello World!";
    }
}
```

For testing this in Lua I have:
```lua
local entityAPI = environment.getEntityAPI();
print(entityAPI)
print(entityAPI.test())
```

This results in:

> table: 1f87cadd
/scan.lua:10: attempt to call field 'test' (a nil value)

So it is definitely returning a table, but still no function. Hopefully I'm doing something simple in the wrong way.

src/main/java/de/srendi/advancedperipherals/common/village/VillagerTrades.java
```diff
@@ -18,14 +19,12 @@ public class VillagerTrades {
 
     @SubscribeEvent
     public static void registerWanderingTrade(WandererTradesEvent event) {
-        if (APConfig.WORLD_CONFIG.enableWanderingTraderTrades.get()) {
```
zyxkad: You added it, you removed it. 🤔

SirEndii:
I think he solved the conflicts a bit wrong there

src/main/java/de/srendi/advancedperipherals/common/village/VillagerTrades.java
lonevox: Lmao I'll fix that

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
```diff
@@ -30,15 +37,17 @@
 import java.util.HashMap;
 import java.util.List;
 import java.util.Map;
+import java.util.UUID;
 import java.util.function.Supplier;
 import java.util.stream.Stream;
 
 public class LuaConverter {
 
-    public static Map<String, Object> entityToLua(Entity entity, boolean detailed) {
```
zyxkad: I kind want keep the `detailed` argument for some forward compatibility

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
```diff
@@ -30,15 +37,17 @@
...
+    public static Map<String, Object> entityToLua(Entity entity) {
         Map<String, Object> data = new HashMap<>();
-        data.put("id", entity.getId());
-        data.put("uuid", entity.getStringUUID());
+        data.put("teamName", ObjectUtil.nullableValue(entity.getTeam(), Team::getName));
```
zyxkad: teamName, inFluid, dimension, isGlowing, isUnderWater, isInLava, and isInWall can counted as detailed information

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
```diff
@@ -52,11 +61,34 @@ public static Map<String, Object> entityToLua(Entity entity, boolean detailed) {
...
+        if (entity.isUsingItem()) {
+            data.put("blocking", entity.isBlocking());
+            Map<String, Object> useItemData = new HashMap<>();
+            useItemData.put("useItem", itemStackToObject(entity.getUseItem()));
+            useItemData.put("usedItemHand", entity.getUsedItemHand().name());
+            useItemData.put("useItemRemainingTicks", entity.getUseItemRemainingTicks());
+            useItemData.put("ticksUsingItem", entity.getTicksUsingItem());
+            data.put("useItemData", useItemData);
+        }
         if (detailed) {
```
zyxkad:
```suggestion
        data.put("scale", entity.getScale());
        data.put("experienceReward", entity.getExperienceReward());
        data.put("armorValue", entity.getArmorValue());
        data.put("mobType", mobTypeToString(entity.getMobType()));
        data.put("sensitiveToWater", entity.isSensitiveToWater());
        if (detailed) {
            data.put("noActionTime", entity.getNoActionTime());
            data.put("lastHurtMob", entityToLuaUUID(entity.getLastHurtMob()));
            data.put("lastHurtByMob", entityToLuaUUID(entity.getLastHurtByMob()));
            data.put("killCredit", entityToLuaUUID(entity.getKillCredit()));
            data.put("lastDamageSource", ObjectUtil.nullableValue(entity.getLastDamageSource(), LuaConverter::damageSourceToObject));
            data.put("arrowCount", entity.getArrowCount());
            data.put("stingerCount", entity.getStingerCount());
            data.put("yHeadRot", entity.getYHeadRot());
            data.put("fallFlyingTicks", entity.getFallFlyingTicks());
            data.put("sleeping", entity.isSleeping());
            if (entity.isUsingItem()) {
                data.put("blocking", entity.isBlocking());
                Map<String, Object> useItemData = new HashMap<>();
                useItemData.put("useItem", itemStackToObject(entity.getUseItem()));
                useItemData.put("usedItemHand", entity.getUsedItemHand().name());
                useItemData.put("useItemRemainingTicks", entity.getUseItemRemainingTicks());
                useItemData.put("ticksUsingItem", entity.getTicksUsingItem());
                data.put("useItemData", useItemData);
            }
```

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
```diff
@@ -67,19 +99,12 @@ public static Map<String, Object> livingEntityToLua(LivingEntity entity, boolean
...
-        if (animal instanceof IForgeShearable shareable && !itemInHand.isEmpty()) {
-            data.put("shareable", shareable.isShearable(itemInHand, animal.level, animal.blockPosition()));
-        }
-        return data;
-    }
-
     public static Map<String, Object> playerToLua(Player player, boolean detailed) {
         Map<String, Object> data = livingEntityToLua(player, detailed);
+        data.put("secondaryUseActive", player.isSecondaryUseActive());
+        data.put("hasContainerOpen", player.hasContainerOpen());
```
zyxkad: what's the usage of `hasContainerOpen` here?

lonevox: It's true if the player has a container open that isn't their inventory. Could be used to detect players opening non-trapped chests. In general I have made as much info available as possible as players will probably find a use for all of it.

zyxkad: imo if you provide `hasContainerOpen` here you should also provide which container the player opened (e.g. give a block position of the inventory, and ignore when player opened any container which is not a block container), because the player can open a container in a massive environment and the value will be kinda useless

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
```diff
@@ -236,6 +309,69 @@ public static Map<String, Object> fluidToObject(@NotNull Fluid fluid) {
...
+
+    public static UUID luaToUUID(Map<?, ?> uuidList) {
+        int msb1 = (int) (double) uuidList.get(1.);
+        int msb2 = (int) (double) uuidList.get(2.);
+        int lsb1 = (int) (double) uuidList.get(3.);
+        int lsb2 = (int) (double) uuidList.get(4.);
+        long msb = (((long) msb1) << 32) | (msb2 & 0xffffffffL);
+        long lsb = (((long) lsb1) << 32) | (lsb2 & 0xffffffffL);
+        return new UUID(msb, lsb);
+    }
```
zyxkad: Well let's be easy here, use string but not int list

zyxkad (changes requested):

src/main/java/de/srendi/advancedperipherals/common/configuration/PeripheralsConfig.java
```diff
@@ -136,7 +137,8 @@ public PeripheralsConfig() {
 
         enablePlayerDetector = builder.comment("Enable the Player Detector or not.").define("enablePlayerDetector", true);
         playerDetMaxRange = builder.comment("The max range of the player detector functions. If anyone use a higher range, the detector will use this max range. -1 for unlimited").defineInRange("playerDetMaxRange", -1, -1, Integer.MAX_VALUE);
-        playerSpy = builder.comment("Activates the \"getPlayerPos\" function of the Player Detector").define("enablePlayerPosFunction", true);
+        enablePlayerPosFunction = builder.comment("Activates the \"getPlayerPos\" function of the Player Detector").define("enablePlayerPosFunction", true);
+        enablePlayerUUIDFunction = builder.comment("Activates the \"getPlayerUUID\" function of the Player Detector").define("enablePlayerUUIDFunction", true);
```
zyxkad: Is there any reason not to get player's UUID but only use their names as identities?

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
lonevox: Getting the real UUID from the UUID string is _slow_. I had noticeable slowdowns during testing because of it. You don't really need to deal with the UUID in Lua land, you just move it around and call methods with it, so it doesn't really matter how it is represented.

src/main/java/de/srendi/advancedperipherals/common/configuration/PeripheralsConfig.java
lonevox: UUIDs are guaranteed to be unique, and the UUID can be used to get additional information from the player via EntityAPI.

zyxkad: I see, but in fact you can search player's UUID by their name from mojang nameserver, so instead of prevent player to get player's UUID in game, we should set limit at entity API.

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
zyxkad: > Getting the real UUID from the UUID string is slow.

I have no idea why parse a UUID string will be slow, UUID basically just a hex sequence, and it should be really fast. But I'll do some test and benchmark later.

> You don't really need to deal with the UUID in Lua land

That's false, sometimes player may want store entities in a table, and use UUID as a key.

src/main/java/de/srendi/advancedperipherals/common/configuration/PeripheralsConfig.java
lonevox: No need to contact mojang servers if the UUID is already available as a field, no?

zyxkad: Yep, so do you think the `enablePlayerUUIDFunction` flag still have any usage?

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
lonevox: I didn't think it would be slow either. There is a chance that the issue was something else and I optimized in the wrong spot here. And thats true, I didnt think of using UUID as a table key.

src/main/java/de/srendi/advancedperipherals/common/configuration/PeripheralsConfig.java
lonevox: Oh I see your point. What do you mean by setting a limit at the entity API?

zyxkad: > Oh I see your point. What do you mean by setting a limit at the entity API?

For example:
```java
if (!config.noPlayerInfo && entity instanceof Player) {
  return playerToLua(xxx)
}
return entityToLua(xxx)
```

lonevox: Ah yes that makes sense, I'll do that.

zyxkad: Or inside playerToLua, do:
```java
    public static Map<String, Object> playerToLua(Player player, boolean detailed) {
        Map<String, Object> data = livingEntityToLua(player, detailed);
        if (config.noPlayerInfo) {
            return data;
        }
        ...
```

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
zyxkad: I just did some benchmark, `UUID.fromString` use average 20ns/op, and parse from map use `10ns/op`.
parse from string is twice slower, but still only slower 10ns, which should not be a big deal I believe

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
lonevox: I can't think of a reliable way to get the block position of the container the player opened. The only thing I can think of is using `Entity.pick` to cast a ray when `playerToLua` is called. The issue is `playerToLua` could be called at any time, and a mob may have walked in front of the player while the player has the container open, which means the block position wouldn't be found.

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
lonevox: Oh it must have been something else then. I'll change it back.

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
zyxkad: Alr, maybe just give the `containerId` as well as `stateId` then

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
```diff
@@ -405,7 +384,8 @@ public static Object effectToObject(MobEffectInstance effect) {
         return map;
     }
 
-    private static Map<Integer, Object> entityToLuaUUID(Entity entity) {
-        return ObjectUtil.nullableValue(entity, e -> LuaConverter.uuidToLua(e.getUUID()));
+    @Nullable
+    private static String entityToUUIDString(@Nullable Entity entity) {
+        return ObjectUtil.nullableValue(entity, e -> e.getUUID().toString());
```
zyxkad: I'm just interested why don't just do `a == null ? null : a.xxx` but pass a callback instead?

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
lonevox: Those aren't unique unfortunately. After talking on the NeoForge discord this can't really be done without large caveats, which would be difficult to explain to the player. Still, I think `hasContainerOpen` could still be used to for example make a sneaky turtle that only moves when the player is distracted looking in blocks. It could also be used in combination with getting the players position to determine what block they are interacting with in a restricted environment like a minigame. I've decided to make `hasContainerOpen` part of the `detailed` info.

src/main/java/de/srendi/advancedperipherals/common/util/LuaConverter.java
lonevox: Uhhhh good question. A little bit of overengineering I think. I'll change those

src/main/java/de/srendi/advancedperipherals/common/configuration/PeripheralsConfig.java
lonevox: I've replaced `enablePlayerUUIDFunction` with `enablePlayerAccess`, which stops players being used in the EntityAPI. There isn't much point in restricting `playerToLua` because you're already getting much more information from the `livingEntityToLua` and `entityToLua` calls that will happen anyway.

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/apis/EntityAPI.java
```diff
@@ -0,0 +1,81 @@
...
+
+    public static final EntityAPI ENTITY_API = new EntityAPI();
+
+    @LuaFunction(mainThread = true)
+    public final Object getNBT(IArguments arguments) throws LuaException {
+        if (!APConfig.API_CONFIG.enableGetNBT.get())
+            throw new LuaException("This function is disabled in the config. Activate it or ask an admin if they can activate it.");
+        Entity entity = EntityUtil.getEntityFromUUID(UUID.fromString(arguments.getString(0)));
+        if (!APConfig.API_CONFIG.enablePlayerAccess.get() && entity instanceof Player)
+            throw new LuaException("Using players in EntityAPI is disabled in the config. Activate it or ask an admin if they can activate it.");
```
zyxkad: Do you think that return `nil, error_msg` is better than throw a `LuaException`? Since when the exception happens the client did not pass an unexpected value.
if we returns the error as a return value script may detect that error easier as well.

lonevox: I think that makes sense for `enablePlayerAccess`, as you might be looping through nearby entities calling an EntityAPI function on them and then it would unexpectedly throw if one of those entities is a player. For the enable function config options, if they are disabled the function truly shouldn't be called at all, so I think throwing an error is fine for those. I'll change `enablePlayerAccess` to return `nil, error_msg`.

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/apis/EntityAPI.java
```diff
@@ -0,0 +1,81 @@
...
+    }
+
+    @LuaFunction(mainThread = true)
+    public final Map<String, Object> getPos(IArguments arguments) throws LuaException {
+        if (!APConfig.API_CONFIG.enableGetPos.get())
+            throw new LuaException("This function is disabled in the config. Activate it or ask an admin if they can activate it.");
+        Entity entity = EntityUtil.getEntityFromUUID(UUID.fromString(arguments.getString(0)));
+        if (!APConfig.API_CONFIG.enablePlayerAccess.get() && entity instanceof Player)
+            throw new LuaException("Using players in EntityAPI is disabled in the config. Activate it or ask an admin if they can activate it.");
+        return LuaConverter.vec3ToLua(entity.getPosition(1));
```
zyxkad:
```suggestion
        return LuaConverter.vec3ToLua(entity.position());
```
I believe you want `position()`, and forge is not good at naming yeah

zyxkad:
Thanks for your contribution! ❤️ 
Everything looks good now. However, I'll still do more client test before merging it, just to be responsible ...
I may finish testing this today, or tomorrow.

Still don't know if query with UUID is a good move, because this change force users do more main thread invokes, it increased the response time, and may decrease the ease of use.

lonevox:
It does decrease ease of use in that you need to manage UUIDs in Lua, but it allows for the `EnvironmentDetectorPeripheral`, `PlayerDetectorPeripheral`, and `AutomataEntityHandPlugin` to use a shared API for getting entity info instead of having duplicate functions.

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/apis/EntityAPI.java
```diff
@@ -0,0 +1,81 @@
...
+     * doesn't overlap with NBT, it ends up mainly being values that exist only at runtime.
+     */
+    @LuaFunction(mainThread = true)
+    public final Object getData(IArguments arguments) throws LuaException {
+        if (!APConfig.API_CONFIG.enableGetData.get())
+            throw new LuaException("This function is disabled in the config. Activate it or ask an admin if they can activate it.");
+        Entity entity = EntityUtil.getEntityFromUUID(UUID.fromString(arguments.getString(0)));
+        if (!APConfig.API_CONFIG.enablePlayerAccess.get() && entity instanceof Player)
+            return MethodResult.of(null, "Using players in EntityAPI is disabled in the config. Activate it or ask an admin if they can activate it.");
+        return LuaConverter.completeEntityToLua(entity);
```
zyxkad: Need a detailed argument here.
```suggestion
        boolean detailed = args.count() > 1 ? args.getBoolean(1) : false;
        if (!APConfig.API_CONFIG.enablePlayerAccess.get() && entity instanceof Player)
            throw new LuaException("Using players in EntityAPI is disabled in the config. Activate it or ask an admin if they can activate it.");
        return LuaConverter.completeEntityToLua(entity, detailed);
```

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/apis/EntityAPI.java
```diff
@@ -0,0 +1,82 @@
...
+     * doesn't overlap with NBT, it ends up mainly being values that exist only at runtime.
+     */
+    @LuaFunction(mainThread = true)
+    public final Object getData(IArguments arguments) throws LuaException {
+        if (!APConfig.API_CONFIG.enableGetData.get())
+            throw new LuaException("This function is disabled in the config. Activate it or ask an admin if they can activate it.");
+        Entity entity = EntityUtil.getEntityFromUUID(UUID.fromString(arguments.getString(0)));
+        boolean detailed = args.count() > 1 ? args.getBoolean(1) : false;
+        if (!APConfig.API_CONFIG.enablePlayerAccess.get() && entity instanceof Player)
+            throw new LuaException("Using players in EntityAPI is disabled in the config. Activate it or ask an admin if they can activate it.");
```
zyxkad:
```suggestion
            return MethodResult.of(null, "Using players in EntityAPI is disabled in the config. Activate it or ask an admin if they can activate it.");
```
I'm dumb I was pasting outdated code

zyxkad:
Ok, so when an entity is not exists any more, the entity API will throw a java error, we should not expect that, instead we should return `nil, "ENTITY_NOT_EXISTS"` or some else message

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/apis/EntityAPI.java
```diff
@@ -0,0 +1,101 @@
...
+    @LuaFunction(mainThread = true)
+    public final Object getNBT(IArguments arguments) throws LuaException {
+        if (!APConfig.API_CONFIG.enableGetNBT.get())
+            throw new LuaException("This function is disabled in the config. Activate it or ask an admin if they can activate it.");
+        try {
+            Entity entity = EntityUtil.getEntityFromUUID(UUID.fromString(arguments.getString(0)));
+            if (!APConfig.API_CONFIG.enablePlayerAccess.get() && entity instanceof Player)
+                return MethodResult.of(null, "Using players in EntityAPI is disabled in the config. Activate it or ask an admin if they can activate it.");
+            return NBTUtil.toLua(entity.serializeNBT());
+        } catch (IllegalArgumentException e) {
```
zyxkad: I see you want catch the `IllegalArgumentException` error that `getEntityFromUUID` throws.
First of all, not only `getEntityFromUUID` throws IllegalArgumentException, there is a lot internal java methods throws it, e.g. `UUID.fromString`
Secondly, as you can see in the first point, try catch makes program process unclear, you will never know what throws and where the exceptions is being throws, even if you defined a specific error `EntityMissingException` you will not know if that exception is formed by the current searching entity missing or if it is because some else entity is missing.

zyxkad: Then, why not just make `getEntityFromUUID` returns `null` when entity was not found? You can check the null value easily as well.

zyxkad (changes requested):

zyxkad:
Hm and with the entity API, you can get any entities data once you know their UUID, no matter where they are. That probably is a little bit too powerful, may be abused, and not what we expect.

I'd ask @SirEndii's opinion for that.

---

Currently two solutions I can think of:
1. Check the entities position & range before returns the data.
2. Snapshot the entities' status when invoking `scanEntity` or such method, search in the snapshot when invoking getData.

However both of them will make EntityAPI depends on the peripheral's status again

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/apis/EntityAPI.java
lonevox: Sure I'll do that, and add another error check for `UUID.fromString`.

zyxkad: You do not need to check the error of `UUID.fromString`, because that is an explicit error, which users should not pass a non-uuid value

lonevox:
I'm doing slightly weird validation by using Pair, but it is better than before. It would be nice if Java had an `Either` type or Rusts struct-like enums, but we make do.

SirEndii:
> I'd ask @SirEndii's opinion for that.

I would prefer your option one, seems to be the simplest, and it would follow the same logic as with other functions and peripherals

lonevox:
I'm currently opting for having `EntityAPI` be constructed by `EnvironmentDetectorPeripheral::getEntityAPI` like so (it used to return a static `EntityAPI` instance):
```java
@LuaFunction
public final EntityAPI getEntityAPI() {
    return new EntityAPI(getPeripheralOwner());
}
```
This means that within the `EntityAPI` it can check the distance from the entity to the peripheral whenever a function is called. However, I'm unsure of what range the `EntityAPI` should have, and how to handle range. It could just be a flat max distance value such as 5 blocks, or it could be a `SphereOperation`. If the right way to go is a `SphereOperation`, then here's some options I'm considering:

1. I could add a new `SphereOperation` value for each `EntityAPI` function (as `EntityAPI::getPos` should probably be cheaper than `EntityAPI::getNBT` based on how much less data is returned) but this would pollute `SphereOperation`, as it currently has 2 values and I would be adding 6 more.
2. I could add just one `SphereOperation` value called `GET_ENTITY_DATA` and have the free radius the same as the cost radius so it is a free operation. The idea behind it being free is the player already needed to run `EnvironmentDetectorPeripheral::scanEntities` to get the entity UUIDs, however this might as well be free because you can now use those UUIDs to call `EntityAPI` functions for that entity forever for no extra cost.
3. A compromise is to add 2 `SphereOperation` values:

    - `GET_SMALL_ENTITY_DATA` (for `getName`, `getPos`, `getBoundingBox`, and `getPersistentData`).
    - `GET_LARGE_ENTITY_DATA` (for `getNBT` and `getData`).

Or `SphereOperation` might not be the way to do this at all. I'd like to know what you guys think.

There is also a whole other issue of the fact that a malicious player could get their hands on a victims `EntityAPI` instance (since they're instanced by `EnvironmentDetectorPeripheral` now) and call expensive functions on it from anywhere. This lends itself to unfortunately ditching the `EntityAPI` and moving the functions to `EnvironmentDetectorPeripheral`, which is becoming a more attractive option.

SirEndii:
Sorry for not answering, I completely forgot this PR

I would prefer the use of SphereOperation, your second option sounds more reasonable to me
> There is also a whole other issue of the fact that a malicious player could get their hands on a victims EntityAPI instance (since they're instanced by EnvironmentDetectorPeripheral now) 

If the function creates a new Entity API for every `getEntityAPI()` function call, wouldn't a user need to save the entity api inside a local field so the cooldowns and other stuff are stored? If a user always creates a new EntityAPI instance via the function, wouldn't they also create new cooldowns/reset them,

Maybe you're right, and the best way is to move all the functions into the EnvDetector itself. I at least would now prefer that

SirEndii:
Greetings @lonevox 
I would just like to know how things are going for you in terms of the time you have and what else you are planning for the PR

zyxkad:
I just have an idea that you should not try to use UUID to index entity data at all, you can just returns a table include a method that takes no argument for extra information, such as:
```lua
{
  id = x,
  uuid = xx,
  name = xxx,
  type = xxxx,
  getAdvancedInfo = function() return <wrappedJavaMap> end,
  getNBT = function() return <convertedNBTData> end,
}
```

PR #554, opened 2024-01-27
Add random error to getPlayerPos

qcwrenshu:

Feature

Calling `PlayerDetectorPeripheral.getPlayerPos` currently returns a player's exact position regardless of how far they are from the player detector.
As administrator of a server that uses this mod, I wanted to retain the ability to get a player's general position in the world, but I also wanted to maintain player privacy (such as avoiding exposing hidden bases) by not returning an exact position.

This PR introduces random error to X, Y, and Z dimensions of a player's position that varies based on how far the player is from the player detector. The error multiplier increases sub-linearly based on Euclidean distance up to a point.

    Values are currently set in the code as follows:
     - **Min distance** (under this distance, error is never added): 50 (blocks/meters)
     - **Max distance** (the distance at which max error is added): 10000 (blocks/meters)
     - **Max error** (the max error added to each axis): ±2500 (blocks/meters)

    These values could/should probably be changed or added as configuration options.

    This PR also introduces a configuration option `enablePlayerPosRandomError` that enables or disables adding random error to the player's position. This configuration option is set to false by default to avoid a breaking change.

No. The feature introduced in this PR is disabled by default.

* **Other information**:
The numeric values in this change (such as min/max distance and max error) are currently hard-coded in the code and should probably be added to configuration.
Consider adding error to player yaw, pitch, and especially eyeHeight.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/PlayerDetectorPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/PlayerDetectorPeripheral.java
@@ -213,23 +213,37 @@ public class PlayerDetectorPeripheral extends BasePeripheral<IPeripheralOwner> {
 
         Map<String, Object> info = new HashMap<>();
 
-        double x = existingPlayer.getX();
-        double y = existingPlayer.getY();
-        double z = existingPlayer.getZ();
+        double x = existingPlayer.getX(), y = existingPlayer.getY(), z = existingPlayer.getZ();
 
         if (APConfig.PERIPHERALS_CONFIG.playerSpyRandError.get()) {
-            double distance = Math.sqrt(Math.pow(x - getPos().getX(), 2) + Math.pow(y - getPos().getY(), 2) + Math.pow(z - getPos().getZ(), 2));
+            // We apply random error to the returned player position if so enabled in the configuration.
 
-            final int minDistance = 50;
-            final int maxDistance = 10000;
-            final int maxError = 2500;
+            // minDistance: Below this distance, the player's exact position is returned
+            int minDistance = APConfig.PERIPHERALS_CONFIG.playerSpyPreciseMaxRange.get();
+            // maxError: The maximum amount of blocks that the player's position can be off by (on each axis) at the max distance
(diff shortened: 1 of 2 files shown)


src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/PlayerDetectorPeripheral.java
```diff
@@ -212,9 +212,30 @@ public final Map<String, Object> getPlayerPos(String username) throws LuaExcepti
...
+
+        double x = existingPlayer.getX();
+        double y = existingPlayer.getY();
+        double z = existingPlayer.getZ();
+
+        if (APConfig.PERIPHERALS_CONFIG.playerSpyRandError.get()) {
+            double distance = Math.sqrt(Math.pow(x - getPos().getX(), 2) + Math.pow(y - getPos().getY(), 2) + Math.pow(z - getPos().getZ(), 2));
+
+            final int minDistance = 50;
+            final int maxDistance = 10000;
```
zyxkad: Just use the `playerDetMaxRange` in config?

zyxkad:
PS: we do not use ✅ or ❌. You may want use

```markdown
- [x] something
```
Instead of ✅
- [x] something

And just do
```markdown
- [ ] something else
```
Instead of ❌
- [ ] something else

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/PlayerDetectorPeripheral.java
SirEndii: I am not 100% sure what the purpose of these variables are. It seems that the error shouldn't go above 2500 if the player is further than 10000 blocks away

The code lacks a bit of documentation/comments

SirEndii (changes requested):
There are some checkstyle issues. Either run `gradlew build` in your dev environment to see them or check the teamcity log

qcwrenshu:
Apologies. I've documented the code and fixed checkstyle issues. I've also added configuration options for 'max distance' and 'max error'.

SirEndii (approved):

SirEndii:
Looks good to me!
Thank you for your contribution, I will try to add documentation for the changes before I will release them

PR #562, opened 2024-02-23
Add game and junit tests to AP

SirEndii:
This PR aims to add the long needed game tests to AP. It's still a draft.
We will use the Testing Framework from CC to accomplish this
The framework is implemented, and I started to work on the first tests

- Checklist
- [ ] - Add JUnit tests for common AP functions
- [ ] - Mod Integrations
  - <details>
  
    - [x] - Botania
    - [x] - AE 2
    - [x] - Note Blocks
    - [x] - Beacons
    - [ ] - RS (currently not possible)
    - [ ] - Create
    - [ ] - Mekanism
    - [ ] - Powah
    - [ ] - Minecolonies (currently not possible)
  </details>
- [ ] - Peripherals
  - <details>
  
    - [ ] - Environment Detector
    - [x] - Chat Box
    - [ ] - Player Detector
    - [x] - Energy Detector
    - [ ] - Inventory Manager
    - [x] - Redstone Integrator
    - [x] - Block Reader
    - [x] - Geo Scanner
    - [x] - NBT Storage
  </details>
- [ ] - Turtles
  - <details>
  
    - [ ] - Chatty
    - [ ] - Chunky (Not possible to test)
    - [ ] - Compass
    - [ ] - Player Detector
    - [ ] - Environment Detector
    - [ ] - Geo
    - [ ] - Weak Automata
    - [ ] - OP Weak Automata
    - [ ] - Husbandry Automata
    - [ ] - OP Husbandry Automata
    - [ ] - End Automata
    - [ ] - OP End Automata
  </details>
- [ ] - Pocket Computers (?) Currently not possible without an entity that loads the pocket
  - <details>
  
    - [ ] - Chatty
    - [ ] - Player Detector
    - [ ] - Enviroment
    - [ ] - Geo
  </details>
- [x] - Documentation 

ToDos:
- [ ] The [Geo Scanner docs](https://docs.advanced-peripherals.de/peripherals/geo_scanner/) list a `getScanCooldown` method which currently does not exist in AP, so I cannot test it (the test currently uses the "internal" `getOperationCooldown("scanBlocks")` method)
- [x] The `isOnEnchantedSoil` function on Botania Mana Flowers always returns false, even when the flower is placed on enchanted soil. This seems to be the case because the `overgrowth` variable on a Mana Flower is only set to true during the flower tick (see https://github.com/VazkiiMods/Botania/blob/1.20.x/Xplat/src/main/java/vazkii/botania/api/block_entity/SpecialFlowerBlockEntity.java#L100)
- [ ] The `isEmpty` method on Botania Mana Pools is currently missing
- [ ] For the Botania Mod Integration, some methods are only available in the 1.20.1 versions of AP, so they're currently commented out in the .lua test files. When merging this to the 1.20.1 branch, these methods should be uncommented
- [ ] Botania Mana Spreaders have a `getBounding` method returning the coordinates of the block that the spreader is directed towards, shouldn't it be called `getBinding` then?
- [ ] The `isEmpty` function on Botania Mana Spreaders does not return if the mana in the spreader is empty but if the item handler of the spreader is empty
- [ ] On Note Blocks, the method `changeNoteBy` is named somewhat weird. It sets the note value to the given parameter and does not change the note by x amount. Shouldn't it be called `setNote` or something?


SirEndii:
I will mark this as high priority. I will work on this asap since we really need tests for AP

SirEndii:
I've implemented (copied) the needed stuff from the test source sets, but there is currently another issue with hamcrest and junit I need to work on
I'm going to fix this asap. For now there are the cc tests in AP for testing purposes

SirEndii:
I fixed everything that need to be fixed (I think)
The game tests are now ready to be created

build.gradle
```diff
@@ -126,38 +148,69 @@ minecraft {
...
             workingDirectory project.file('test-files/client')
             parent runs.client
 
+            def old = lazyTokens["minecraft_classpath"]
+            lazyToken("minecraft_classpath") {
+                // Add all files in testMinecraftLibrary to the classpath.
+                def allFiles = new HashSet<String>();
+
+                def oldVal = old?.get()
+                if (!oldVal != null && !oldVal.isEmpty()) allFiles.addAll(oldVal.split(File.pathSeparatorChar))
```
zyxkad: what is `!oldVal != null`

src/testMod/java/de/srendi/advancedperipheralstest/TestGameTests.java
```diff
@@ -0,0 +1,19 @@
+package de.srendi.advancedperipheralstest;
```
zyxkad:
```suggestion
package de.srendi.advancedperipherals.test;
```

src/testMod/kotlin/dan200/computercraft/gametest/Computer_Test.kt
```diff
@@ -0,0 +1,72 @@
+/*
```
zyxkad: Is these gametests from CC:T really necessary? Is our test depending on it?

src/testMod/resources/data/advancedperipheralstest/computer/tests/craftos_test.sends_basic_rednet_messages.echo.lua
```diff
@@ -0,0 +1,8 @@
+os.getComputerID = function() return 1 end
```
zyxkad: I don't think we need to test CC:T stuffs in AP.

SirEndii: Yeah these are just for testing. I used them to see if everything works fine in the testing environment

build.gradle
SirEndii: That's the minecraft classpath
The value of the returned supplier can be null, so we check that

build.gradle
```diff
@@ -126,38 +148,69 @@ minecraft {
...
             workingDirectory project.file('test-files/client')
             parent runs.client
 
+            def oldClasspath = lazyTokens["minecraft_classpath"]
+            lazyToken("minecraft_classpath") {
+                // Add all files in testMinecraftLibrary to the classpath.
+                def allFiles = new HashSet<String>();
+
+                def minecraftClasspath = oldClasspath?.get()
+                if (!minecraftClasspath != null && !minecraftClasspath.isEmpty()) allFiles.addAll(minecraftClasspath.split(File.pathSeparatorChar))
```
zyxkad: Well, I mean, is **`!`** `minecraftClasspath != null` a correct syntax? I checked gradle docs and the syntax should same as java?

SirEndii: Ahhh that's why intellij showed me a warning 
Didn't saw that, oops

SirEndii:
We're on a good run.
I've added the first game tests using an ingame script. I will see if I can run the functions via the kotlin script like CC does it with the APIs like turtle.*. But I doubt that this is easily possible with peripherals

We just now need to add the game tests itself. I will also write a documentation for that

SirEndii:
Had another error in my thinking. gradle build already executes test and runGameTestServer, I'll change it later

SirEndii:
I think there is nothing left to do in terms of the framework. Now we just need to create the tests. There is a me bridge tests to test something a bit more complex and the environment detector which currently just tests one function.

I've created a documentation markdown file located in docs/CREATING_TESTS.md. It explains the basics, how to run and write tests. But it needs some more love with examples and code.
If someone wants to create tests, feel free to do so😅  Would speed things up
For non collaborators, you could create a PR which targets the feat/tests branch. But I am not too sure how well this goes

src/testMod/resources/data/advancedperipheralstest/computer/tests/peripheraltest.environment.lua
```diff
@@ -0,0 +1,5 @@
+detector = peripheral.find("environmentDetector")
+test.eq(detector ~= nil, true, "Peripheral not found")
```
zyxkad: not sure if the method exists but I believe `test.neq(detector, nil` will better than `test.eq(x ~= nil`

src/testMod/resources/data/advancedperipheralstest/computer/tests/peripheraltest.environment.lua
```diff
@@ -0,0 +1,5 @@
...
+
+isRaining = detector.isRaining()
+test.eq(false,isRaining, "It should not rain")
```
zyxkad:
```suggestion
test.eq(false, isRaining, "It should not rain")
```

src/testMod/resources/data/advancedperipheralstest/computer/tests/peripheraltest.mecrafting.lua
```diff
@@ -0,0 +1,30 @@
+sleep(4)
+bridge = peripheral.wrap("bottom")
+test.assert(bridge, "Peripheral not found")
+
+isOnline = bridge.isConnected()
+test.assert(isOnline, "Bridge is not connected/system is not online")
+
+validEnergy = bridge.getEnergyUsage() > 24
+test.assert(validEnergy, tostring(bridge.getEnergyUsage()) .. "Consumption does not seem right")
```
zyxkad:
```suggestion
test.assert(validEnergy, tostring(bridge.getEnergyUsage()) .. " Consumption does not seem right")
```

or `string.format` will looks better than string join

src/testMod/resources/data/advancedperipheralstest/computer/tests/peripheraltest.mecrafting.lua
```diff
@@ -0,0 +1,30 @@
...
+test.assert(not isItemCrafting, "There shouldn't be a crafting job")
+
+-- Should be true. We have a pattern for it.
+isItemCraftable = bridge.isItemCraftable(stickFilter)
+test.assert(isItemCraftable, "Stick should be craftable")
+
+-- Logs should not be craftable, we don't have a pattern for it
+isItemCraftable = bridge.getItem(logFilter).isCraftable
+-- We use == false here since the not keyword would also return true if the value is nil
+test.assert(isItemCraftable == false, "Log should not be craftable")
```
zyxkad, line 34:
```suggestion
test.assert(not isItemCraftable, "Log should not be craftable")
```

src/testMod/resources/data/advancedperipheralstest/computer/tests/peripheraltest.mecrafting.lua
```diff
@@ -0,0 +1,30 @@
...
+
+-- Planks are craftable
+isItemCraftable = bridge.getItem(planksFilter).isCraftable
+test.assert(isItemCraftable == true, "Planks should be craftable")
```
zyxkad, line 38:
```suggestion
test.assert(isItemCraftable, "Planks should be craftable")
```

it's better also keep same style when writing lua scripts, either all of them use `xxx == true`, `xxx == false` or all of them use `xxx`, `not xxx`

src/testMod/resources/data/advancedperipheralstest/computer/tests/peripheraltest.mecrafting.lua
SirEndii: There is a comment above that line which explains why I did that

src/testMod/resources/data/advancedperipheralstest/computer/tests/peripheraltest.mecrafting.lua
SirEndii: Yeah you're right
But sometimes it does not really make sense to use == true/false
And sometimes I just need to use == true/false like in line 25

zyxkad:
by the way why is chunky turtle not possible to test?
I think the test cases are:

- load when server start:
   1. place turtle to (edge of/far away in) the world
   2. restart server
   3. check if the turtle id exists, and/or check if the chunk is loaded
- unload chunk when moving/destroyed:
   1. place turtle to far away in the world
   2. make the turtle move across the border, or destroy the turtle
   3. check if the back chunk is unloaded
- load chunk when moving:
   1. place turtle to far away in the world
   2. make the turtle move across the border
   3. check if the front/side chunk is loaded

I do not know if that's possible to do with the current test framework.

SirEndii:
Okay you are right, not very complicated

SirEndii:
I currently just tend to merge this, even unfinished. Imo testing is important, especially with all the new changes. But I don't have the time to implement ALL of that.
I will change my to-do list a bit and sort it by priority. Something which is annoying to test by hand should have the highest prio and vice versa

PR #570, opened 2024-03-28
Fix meBridge.listCraftableItems and meBridge.listCraftableFluids to not be quadratic.

tomprince:
Applied Energistics will return the set of all craftable things. Iterate over that to generate the list of craftable things, instead of something more complicated.

Previously, the code would

1. Iterate over *every* item in the entire ME system.
2. For *each* non-craftable item, iterate over every craftable item and a) create the corresponding lua object b) if the craftable item isn't present in ME system, check if the list of objects to return contains that object (which requires sequentially scanning the list of objects to return.

This meant that

1. If there were no non-craftable items in the ME system, none of the craftable items that weren't present in the system would be returned.
2. The code was quadratic or cubic in runtime.

---

- [X] The commit message are well described
- [ ] All changes have fully been tested

This fixes a couple of bugs in listing craftable items/fluids in an ME system.

- Quadratic or cubic runtime for listing craftable items in an ME system.
- Not listing craftable items that are not in the ME system, if there are *no* non-craftable items in the system.

This might change the order of the list of craftable items returned.

* **Other information**:

This PR is against the 1.18 branch, since the modpack I am currently playing is a 1.18 modpack. I've not tested this change against newer versions.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
@@ -75,36 +75,46 @@ public class AppEngApi {
         return null;
     }
 
-    public static List<Object> listStacks(MEStorage monitor, ICraftingService service, int flag) {
+    public static List<Object> listStacks(MEStorage monitor, ICraftingService service) {
         List<Object> items = new ArrayList<>();
         KeyCounter keyCounter = monitor.getAvailableStacks();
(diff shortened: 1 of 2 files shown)


tomprince:
This takes calling `meBridge.listCraftableItems()` for me in my current modpack from several plus seconds to basically instant. I have 777 items in my ME system, and 512 craftable items (of which 351 are not present in the ME system).

SirEndii:
Thank you for the PR!
I will test this a bit deeper and since it's for 1.18 port it to newer versions.

SirEndii (approved):

PR #575, opened 2024-04-06
fix: ME Bridge: Add support for child classes created by other mods

iTrooz:

- [x] (not needed) Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/SirEndii/Advanced-Peripherals-Documentation/pulls). This is not mandatory

When adding some methods for AE2 network usage, https://github.com/IntelligenceModding/AdvancedPeripherals/pull/404 seems to not have taken into account portable cells. Right now, these cells can be put in a AE2 Drive, but won't be counted by methods like `getTotalItemStorage()`

Looking around the source, it seems like `IBasicCellItem` is a common interface for classical and portable cells, so I used it instead.

See above

See above

No

* **Other information**:

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AppEngApi.java
@@ -339,7 +339,7 @@ public class AppEngApi {
                 } else if (APAddons.aeAdditionsLoaded && (stack.getItem() instanceof StorageCell storageCell)) {
                     if (storageCell.getKeyType() != AEKeyType.items())
                         continue;
-                    total += storageCell.getKiloBytes() * 1024;
+                    total += storageCell.getKiloBytes() * 1024L;
                 }
             }
         }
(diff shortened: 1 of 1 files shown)


iTrooz:
Oops sorry, I targeted 1.20 instead of 1.19. Want me to create another PR ? You might need to cherry-pick to the other branch in any case tho

iTrooz:
After more investigation I found another problem that affected me: the "ME Extended Drive" from ExtendedAE, which is a child class of DriveBlockEntity, wasn't read.

I also fixed this in this PR, and renamed it accordingly

In the long term, a better solution would be to directly integrate with ExtendedAE since there are other stuff that doesn't work (for example their "ME Tag Storage Bus", which isn't a child class of anything interesting at first glance, except maybe IStorageProvider)

SirEndii:
> Oops sorry, I targeted 1.20 instead of 1.19. Want me to create another PR ? You might need to cherry-pick to the other branch in any case tho

It's okay for now, not a big deal
But targeting 1.19 the next time would be nice

Your PR looks solid, I'm gonna test this a bit in the course of the day

PR #579, opened 2024-04-11
Migrate to Spotless from Checkstyle

SirEndii:
Advanced Peripherals currently uses the gradle [checkstyle](https://docs.gradle.org/current/userguide/checkstyle_plugin.html) plugin to be able to maintain a uniform code style in the project.

But there are two main reasons why I want to switch from checkstyle to spotless.
- Spotless has way more features
   Spotless has a wider range of features. One among them is the `spotlessApply` task to just fix the style issues automatically while maintaining license headers.
- Advanced custom steps
  If I need custom steps, I can create custom steps using `FormatterStep` and `FormatterFunc` using spotless, which also works for the `spotlessApply` task. Checkstyle "only" allows custom regex checks. Which isn't bad, but I just can do more with these java based custom steps.
  
  
Currently, there aren't any custom steps. The eclipse formatter used provides a decent code style I really like. But I will consider adding custom steps for our annotation and one-liner if statement guidelines


SirEndii:
Currently waiting for any feedback on https://github.com/diffplug/spotless/issues/2095

I am unable to use https://github.com/diffplug/spotless/blob/main/plugin-gradle/README.md#retroactively-slurp-years-from-git-history

PR #581, opened 2024-04-14
Fix NullPointerException when server ticking

zyxkad:

Some mod may call `ServerWorker.tick` with null value, this PR ensure null value never go to the queue, and if it did somehow, the null will never be executed

No

* **Other information**:

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/util/ServerWorker.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/util/ServerWorker.java
@@ -5,13 +5,13 @@ import net.minecraftforge.event.TickEvent;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 import net.minecraftforge.fml.common.Mod;
 
-import java.util.ArrayDeque;
 import java.util.Queue;
+import java.util.concurrent.ConcurrentLinkedQueue;
 
 @Mod.EventBusSubscriber(modid = AdvancedPeripherals.MOD_ID)
 public class ServerWorker {
 
-    private static final Queue<Runnable> callQueue = new ArrayDeque<>();
+    private static final Queue<Runnable> callQueue = new ConcurrentLinkedQueue<>();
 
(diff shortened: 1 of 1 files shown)


SquidDev:
What bug is this trying to fix? I don't see [any usages of ServerWorker.add](https://github.com/search?q=repo%3AIntelligenceModding%2FAdvancedPeripherals%20ServerWorker.add&type=code) that enqueue `null`.

zyxkad:
https://discord.com/channels/734726882058174486/779400553813377074/1229036290264010782

SquidDev:
[That's not very helpful!](https://github.com/IntelligenceModding/AdvancedPeripherals/assets/4346137/431fb81e-f0ef-4b8d-80b8-6942223d3bab). I suspect the issue here is actually due to the non-thread-safe usage of `ArrayQueue`. This should either use a concurrent queue, or obtain a lock on the queue.

zyxkad:
Ok then:
[Error Log](https://github.com/IntelligenceModding/AdvancedPeripherals/files/14971078/message.txt)

zyxkad:
> I suspect the issue here is actually due to the non-thread-safe usage of `ArrayQueue`. This should either use a concurrent queue, or obtain a lock on the queue.

You are right, they are running concurrently

19PHOBOSS98:
merge please : )

zyxkad:
@19PHOBOSS98 can you build your own version based on this PR and test if that error never happen again?

Cause it's not really easy for me to reproduce

SirEndii (approved):

19PHOBOSS98:
> @19PHOBOSS98 can you build your own version based on this PR and test if that error never happen again?
> 
> Cause it's not really easy for me to reproduce

sure thing! I'll be back with what I find

19PHOBOSS98:
it doesn't crash anymore when I tried setting the threads to 4 with all of my drones flying with VS2. Thanks again! Hope this releases soon

PR #583, opened 2024-04-14
Fix NullPointerException when server ticking

zyxkad:
#581 
Critical fix back to 1.18

* **Other information**:
See more at #581

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/util/ServerWorker.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/util/ServerWorker.java
@@ -5,23 +5,28 @@ import net.minecraftforge.event.TickEvent;
 import net.minecraftforge.eventbus.api.SubscribeEvent;
 import net.minecraftforge.fml.common.Mod;
 
-import java.util.ArrayDeque;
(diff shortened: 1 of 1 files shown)


src/main/java/de/srendi/advancedperipherals/common/util/ServerWorker.java
```diff
@@ -5,23 +5,28 @@
...
+        if (call != null) {
+            callQueue.add(call);
+        }
     }
 
     @SubscribeEvent
     public static void serverTick(TickEvent.ServerTickEvent event) {
         if (event.phase == TickEvent.Phase.END) {
-            while (!callQueue.isEmpty()) {
+            while (true) {
```
SirEndii, line 25: Is there a reason why you switched from `while (!callQueue.isEmpty()) {` to `while (true) {`

zyxkad: Because when `poll()` returns `null`, the queue is empty, just want take a less step of check

SirEndii (approved):

PR #622, opened 2024-06-01
Backported changes from the CC: Tweaked Gametest framework present in 1.20.x

dogboy21:
**PLEASE READ THE [GUIDELINES](https://github.com/SirEndii/AdvancedPeripherals/blob/dev/1.20.1/CONTRIBUTING.md) BEFORE MAKING A CONTRIBUTION**

* **Please check if the PR fulfills these requirements**
- [x] The commit message are well described
- [x] Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/SirEndii/Advanced-Peripherals-Documentation/pulls). This is not mandatory
- [x] All changes have fully been tested

* **What kind of change does this PR introduce?** (Bug fix, feature, ...)

Feature

* **What is the current behavior?** (You can also link to an open issue here)

Currently, the used gametest framework in #562 is based on the 1.19.x branch of CC: Tweaked.

* **What is the new behavior (if this is a feature change)?**

Apparently some stuff has changed in the 1.20.x branch (for example support for client gametests) which this PR aims to backport to AP.

* **Does this PR introduce a breaking change?** (What changes might users need to make in their scripts due to this PR?)

All the old tests run without a problem

PR #628, opened 2024-06-03
Added client test for the Chat Box peripheral

dogboy21:

Feature

No tests for the Chat Box peripheral in #562 

Added a test for the Chat Box with the new client test feature introduced in #622 

No

---

ToDo:

- [x] Implement tests for the`chat` event
- [x] Fix the `runTestClient` and `runGameTestClient` gradle tasks
  - The `runTestClient` task is used to build the test structures, so it needs to have the property `advancedperipheralstest.tags` set to `common,client` so that both server and client tests can be built and executed.
  - The `runGameTestClient` is used to automatically run client tests, so the property `advancedperipheralstest.tags` needs to set to `client` so we only run the client tests, not the server tests.
  - With the current implementation this does not work correcly and both server and client tests are run on the client


SirEndii:
I maybe can look into some of your to-dos when I find time

PR #643, opened 2024-07-09
rename getEnergy -> getStoredEnergy

### LINKED ISSUE #642: Powah Ender cell numeric clamping
### Describe

The Powah mod makes it capable to store insane amounts of power, using `Ender Cells`

The Powah manual states that an Energy Cell can hold up to 9E of power (Still not sure what the unit of measurement `E` is yet)

However, the mod compatibility seems to clamp the read energy stored as an int.

```lua
cell = peripheral.find("enderCell")

cell.getEnergy()
-- Returns "2147483647" (Max 32bit)

cell.getMaxEnergy()
-- Returns "2000000000000" >32 bit
```

As seen above though, the "MaxEnergy" is read correctly, at _2 trillion_

I tested `type(cell.getMaxEnergy())` if it was actually being handled as a string, but it does say `number`.

### Steps to reproduce

1. Obtain an `Ender Cell` and 2x `Energy Cell (Nitro)` (Capable of storing 2B power each)
2. Place the Ender Cell, open the Ender Cell and shift click both Energy Cells inside (Setting capacity at 4B)
3. Use a creative power source of some kind to fill the cell past 2.147B / 32-bit int (of 4B capacity)
4. Use a ComputerCraft computer to attempt reading the current stored power

### Multiplayer?

Yes

### Version

1.20.1-0.7.40r (Latest 1.20.1)

NeoForge 47.1.101 Minecraft 1.20.1

### Screenshots or Videos

_No response_

### Crashlog/log

_No response_

zyxkad:

  Bug fix

  Powah's stuff's getEnergy will return an integer since there is a same name method in cc:t

  After rename it will not conflict anymore, and should return a double now

  No

* **Other information**:
   1.19.2 do not have this bug, but I do not know if I still need to add a method called `getStoredEnergy` in 1.19.2

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/owner/BlockEntityPeripheralOwner.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/owner/BlockEntityPeripheralOwner.java
@@ -40,8 +40,7 @@ public class BlockEntityPeripheralOwner<T extends BlockEntity & IPeripheralTileE
                 return name.getString();
             }
         }
-
-        return result;
+        return "";
     }
 
     @NotNull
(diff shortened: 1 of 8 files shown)


SirEndii:
That the check is not successful makes no sense
But I will ignore it

PR #650, opened 2024-07-26
Monitor feature

zyxkad:
**PLEASE READ THE [GUIDELINES](https://github.com/SirEndii/AdvancedPeripherals/blob/dev/1.20.1/CONTRIBUTING.md) BEFORE MAKING A CONTRIBUTION**

* **Please check if the PR fulfills these requirements**
- [x] The commit message are well described
- [ ] Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/SirEndii/Advanced-Peripherals-Documentation/pulls). Feel free to remove this check if you don't need it
- [ ] All changes have fully been tested

* **What kind of change does this PR introduce?** (Bug fix, feature, ...)
   Feature

* **What is the new behavior (if this is a feature change)?**
   - IntelligenceModding/Advanced-Peripherals-Features#6
   - IntelligenceModding/Advanced-Peripherals-Features#12

* **Does this PR introduce a breaking change?** (What changes might users need to make in their scripts due to this PR?)
   No

PR #659, opened 2024-09-23
Keyboard 

SirEndii:
- [ ] Docs have been added / updated

Feature

Added the keyboard item. An item which can be linked to a computer. A user can now right click the item to unlock their mouse, they can now freely type while the typed characters are send to the linked computer. Will also be a module for the smart glasses.

No

* **Other information**:


SirEndii:
So the keyboard works now
I can bound it to a computer, right click it in the air and start typing. Every key stroke, key presses and releases are synced to the terminal

I now want to remove JEI from the keyboard screen and place a little info text at the top

The last thing I need to do is to implement this as a smart glasses module

SirEndii:
I think everything is done now. Mouse and keyboard events are working, I got the little info text rendered, JEI does not render anymore and everything works fine as a glasses module

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
```diff
@@ -0,0 +1,218 @@
...
+        // Make the text a bit smaller on small screens
+        if (screenWidth <= 1080)
+            scale = 1f;
+
+        poseStack.scale(scale, scale, 1);
+        String text = "Press ESC to close the Keyboard Screen.";
+        float textX = (screenWidth / 2f - minecraft.font.width(text) * scale / 2f) / scale;
+        minecraft.font.drawShadow(poseStack, text, textX, 1, 0xFFFFFF);
+
+        // Prevents JEI/REI/EMI from rendering. Maybe not the best way, but it works for now.
```
zyxkad: Would you mind explain a bit what are you doing here to prevent them render? Or I mean I want to know if this won't disable other mods that decides to constant render something on the screen, such as inventory hud+

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
```diff
@@ -0,0 +1,218 @@
...
+    protected void renderBg(@NotNull PoseStack matrixStack, float partialTicks, int x, int y) {
+    }
+
+    @Override
+    public void renderBackground(@NotNull PoseStack pPoseStack) {
+    }
+
+    @Override
+    public boolean charTyped(char ch, int modifiers) {
+        if (ch >= 32 && ch <= 126 || ch >= 160 && ch <= 255) { // printable chars in byte range
```
zyxkad: ASCII char has a maximum value of `0x7f` (127) btw, not `0xff` or 255, above 0x7f will be considers as a two or more bytes UTF8

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
```diff
@@ -0,0 +1,218 @@
...
+                            clipboard = clipboard.substring(0, newLineIndex1);
+                        } else if (newLineIndex2 >= 0) {
+                            clipboard = clipboard.substring(0, newLineIndex2);
+                        }
+
+                        // Filter the string
+                        clipboard = SharedConstants.filterText(clipboard);
+                        if (!clipboard.isEmpty()) {
+                            // Clip to 512 characters and queue the event
+                            if (clipboard.length() > 512) clipboard = clipboard.substring(0, 512);
```
zyxkad: I guess we better clip before check anything should we not?

zyxkad (commented):
Overall looks good, just have some detail need to deal with

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
SirEndii: I have actually no clue if it also disables other ui addon stuff
So I just render an empty string on the far lower right and jei thinks it does not have space to render
There is maybe a better way, maybe using the JEI API

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
SirEndii: Everything below charTyped is borrowed from CCs screen itself, so I don't know why squid did what 😅

SirEndii:
Sorry to annoy you @SquidDev, just asking you because of https://github.com/IntelligenceModding/AdvancedPeripherals/pull/650#issuecomment-2254387033

I used a piece of the TerminalWidget implementation for our keyboard screen. Just asking if this is okay
See
https://github.com/IntelligenceModding/AdvancedPeripherals/pull/659/files#diff-6374267185184dd9bdd6868cb47752c688a105088d0063d0c5bf7992b4e10ec9R19

And maybe you can answer the two comments related to that from zyxkad

SquidDev:
It might be easier to copy what [`NoTermComputerScreen` does](https://github.com/cc-tweaked/CC-Tweaked/blob/c271ed7c7f9c9e3e927a0939fafb20690b8d40f3/projects/common/src/client/java/dan200/computercraft/client/gui/NoTermComputerScreen.java), and just create a `Terminal` but never render it. Plethora's keyboard UI [actually just subclasses `NoTermComputerScreen` directly](https://github.com/SwitchCraftCC/Plethora-Fabric/blob/1.20.1/src/main/java/io/sc3/plethora/gameplay/modules/keyboard/KeyboardComputerScreen.kt).

SirEndii:
> It might be easier to copy what [`NoTermComputerScreen` does](https://github.com/cc-tweaked/CC-Tweaked/blob/c271ed7c7f9c9e3e927a0939fafb20690b8d40f3/projects/common/src/client/java/dan200/computercraft/client/gui/NoTermComputerScreen.java), and just create a `Terminal` but never render it. Plethora's keyboard UI [actually just subclasses `NoTermComputerScreen` directly](https://github.com/SwitchCraftCC/Plethora-Fabric/blob/1.20.1/src/main/java/io/sc3/plethora/gameplay/modules/keyboard/KeyboardComputerScreen.kt).

Thank you, I didn't see that
That would be simpler

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
SirEndii: I simplified that
I now just define a huge screen image which tells minecraft that the actual screen which we don't need to render is huge
That prevents jei from rendering

build.gradle
```diff
@@ -324,11 +324,12 @@ dependencies {
     compileOnly fg.deobf("com.ldtteam:domum_ornamentum:${domumornamentum_version}:universal")
     compileOnly fg.deobf("com.ldtteam:blockui:${blockui_version}")
     // IMPORTANT. This should be removed/commented when running `runData`
-    runtimeOnly fg.deobf("com.ldtteam:minecolonies:${minecolonies_version}")
+    /*runtimeOnly fg.deobf("com.ldtteam:minecolonies:${minecolonies_version}")
```
zyxkad: I pretend I didn't see those

(although I think you should merge from dev/0.8)

src/main/java/de/srendi/advancedperipherals/client/ClientRegistry.java
```diff
@@ -43,6 +44,7 @@ public static void registerModels(ModelEvent.RegisterAdditional event) {
     public static void onClientSetup(FMLClientSetupEvent event) {
         MenuScreens.register(APContainerTypes.INVENTORY_MANAGER_CONTAINER.get(), InventoryManagerScreen::new);
         MenuScreens.register(APContainerTypes.SMART_GLASSES_CONTAINER.get(), SmartGlassesScreen::new);
+        MenuScreens.register(APContainerTypes.KEYBOARD_CONTAINER.get(), KeyboardScreen::new);
```
zyxkad, line 47: keep things sorted!
```suggestion
        MenuScreens.register(APContainerTypes.INVENTORY_MANAGER_CONTAINER.get(), InventoryManagerScreen::new);
        MenuScreens.register(APContainerTypes.KEYBOARD_CONTAINER.get(), KeyboardScreen::new);
        MenuScreens.register(APContainerTypes.SMART_GLASSES_CONTAINER.get(), SmartGlassesScreen::new);
```

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
```diff
@@ -0,0 +1,126 @@
...
+import net.minecraft.world.entity.player.Inventory;
+import org.jetbrains.annotations.NotNull;
+import org.lwjgl.glfw.GLFW;
+
+/**
+ * A simple screen but without any rendering calls. Used to unlock the mouse so we can freely write stuff
+ * <p>
+ * We just create a terminal which is used to forward all the key presses and mouse clicks but we don't render it.
+ */
+public class KeyboardScreen extends BaseScreen<KeyboardContainer> {
```
zyxkad: If you do not extend from `AbstractContainerScreen`, but inherit `Screen` or `GuiComponent`, then jade should not render on it

src/main/java/de/srendi/advancedperipherals/common/container/KeyboardContainer.java
```diff
@@ -0,0 +1,69 @@
...
+import net.minecraft.nbt.CompoundTag;
+import net.minecraft.world.entity.player.Inventory;
+import net.minecraft.world.entity.player.Player;
+import net.minecraft.world.item.ItemStack;
+import net.minecraft.world.level.Level;
+import net.minecraftforge.server.ServerLifecycleHooks;
+import org.jetbrains.annotations.NotNull;
+import org.jetbrains.annotations.Nullable;
+
+public class KeyboardContainer extends BaseContainer implements ComputerMenu {
```
zyxkad, line 23: And so we don't need Keyboard*Container* anymore

src/main/java/de/srendi/advancedperipherals/common/items/MemoryCardItem.java
```diff
@@ -30,9 +30,8 @@ public boolean isEnabled() {
     public void appendHoverText(ItemStack stack, @Nullable Level levelIn, List<Component> tooltip, TooltipFlag flagIn) {
         super.appendHoverText(stack, levelIn, tooltip, flagIn);
         CompoundTag data = stack.getOrCreateTag();
-        // TODO <0.8>: remove the owner name field
-        if (data.contains("ownerId") && data.contains("owner")) {
-            tooltip.add(EnumColor.buildTextComponent(Component.translatable("item.advancedperipherals.tooltip.memory_card.bound", data.getString("owner"))));
+        if (data.contains("ownerId")) {
+            tooltip.add(EnumColor.buildTextComponent(Component.translatable("item.advancedperipherals.tooltip.binding.boundto", data.getString("ownerId"))));
```
zyxkad: alr we are not showing ownerId here, we still want show the owner's name, so we need an extra step to turn the id into `GameProfile`, then get the player's cached name

zyxkad (changes requested):

src/main/java/de/srendi/advancedperipherals/common/items/MemoryCardItem.java
SirEndii: ah yeah
I can use the new cache thing there I made

src/main/java/de/srendi/advancedperipherals/client/screens/KeyboardScreen.java
```diff
@@ -0,0 +1,239 @@
...
+public class KeyboardScreen extends Screen implements MenuAccess<KeyboardContainer> {
+
+    protected final KeyboardContainer keyboardContainer;
+    protected final InputHandler input;
+    private final Terminal terminalData;
+    private WidgetTerminal terminal;
+    private MouseState mouseState = MouseState.RELEASED;
+    private boolean captureMouse;
+    private boolean regrabingMouse;
+    private byte[] lastPosLock = new byte[0];
```
zyxkad:
```suggestion
    private final byte[] lastPosLock = new byte[0];
```
I forgot make this final so it confused me a bit

zyxkad:
Force merge cause 1.19 is not going to be supported.

PR #667, opened 2024-11-21
Add `broadcast` option to the chat box

Niki4tap:
- [ ] The commit message are well described (they are ok, I can amend the description if strictly necessary)

If possible (range, max range are infinite and chat box is multidimensional), we use `broadcastSystemMessage`, instead of sending one to each player.

We send a message to each player individually, no matter the config options.

Described above.

None, internal change.

* **Other information**:
Main motivation for this change is: make the intent more clear, so mods that operate on chat (specifically tested with [mc2discord](https://github.com/DenisD3D/Mc2Discord), mixins only into `broadcastSystemMessage`), can identify the messages and process them.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
@@ -211,8 +211,6 @@ public class ChatBoxPeripheral extends BasePeripheral<IPeripheralOwner> {
     public final MethodResult sendFormattedMessage(@NotNull IArguments arguments) throws LuaException {
         return withChatOperation(ignored -> {
             String message = arguments.getString(0);
-            int maxRange = APConfig.PERIPHERALS_CONFIG.chatBoxMaxRange.get();
-            int range = arguments.optInt(4, -1);
             ResourceKey<Level> dimension = getLevel().dimension();
             MutableComponent component = Component.Serializer.fromJson(message);
             if (component == null) {
@@ -236,6 +234,19 @@ public class ChatBoxPeripheral extends BasePeripheral<IPeripheralOwner> {
                 return MethodResult.of(null, "illegal prefix");
             }
             preparedMessage.append(component);
+
+            int maxRange = APConfig.PERIPHERALS_CONFIG.chatBoxMaxRange.get();
+            int range = arguments.optInt(4, -1);
+            if (
+                APConfig.PERIPHERALS_CONFIG.chatBoxBroadcast.get()
+                    && APConfig.PERIPHERALS_CONFIG.chatBoxMultiDimensional.get()
+                    && maxRange == -1
+                    && range == -1
+            ) {
+                ServerLifecycleHooks.getCurrentServer().getPlayerList().broadcastSystemMessage(preparedMessage, false);
+                return MethodResult.of(true);
+            }
+
             for (ServerPlayer player : ServerLifecycleHooks.getCurrentServer().getPlayerList().getPlayers()) {
                 if (!APConfig.PERIPHERALS_CONFIG.chatBoxMultiDimensional.get() && player.getLevel().dimension() != dimension) {
                     continue;
@@ -252,8 +263,6 @@ public class ChatBoxPeripheral extends BasePeripheral<IPeripheralOwner> {
     public final MethodResult sendMessage(@NotNull IArguments arguments) throws LuaException {
         return withChatOperation(ignored -> {
             String message = arguments.getString(0);
-            int maxRange = APConfig.PERIPHERALS_CONFIG.chatBoxMaxRange.get();
-            int range = arguments.optInt(4, -1);
             ResourceKey<Level> dimension = getLevel().dimension();
             if (checkBrackets(arguments.optString(2))) {
                 return MethodResult.of(null, "incorrect bracket string (e.g. [], {}, <>, ...)");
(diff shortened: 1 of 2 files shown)


zyxkad:
Looks good for me.
Is this a 1.20.1 only feature or if that can be port to 1.19.2?

Niki4tap:
> Is this a 1.20.1 only feature or if that can be port to 1.19.2?

Not sure, tested only on 1.20.1

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
```diff
@@ -125,6 +122,16 @@ public final MethodResult sendFormattedMessage(@NotNull IArguments arguments) th
                     arguments.optString(2, "[]"),
                     StringUtil.convertAndToSectionMark(arguments.optString(3, ""))
             ).append(component);
+
+            int maxRange = APConfig.PERIPHERALS_CONFIG.chatBoxMaxRange.get();
+            int range = arguments.optInt(4, -1);
+            if (range == -1 && maxRange == -1 && APConfig.PERIPHERALS_CONFIG.chatBoxMultiDimensional.get()) {
+                ServerLifecycleHooks.getCurrentServer().getPlayerList().broadcastSystemMessage(preparedMessage, false);
```
zyxkad, line 246: After some thinking, I think its better to add another config that clearly enable to directly use broadcastSystemMessage when maxRange is infinite. so there is no unmatch/randomly changed behaviour when you change the max range.

zyxkad (changes requested):

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
Niki4tap: I don't think it would cause unexpected behavior, but sure, I'm not against a config option.

I'll add it and rebase the PR as soon as I get a response here: https://github.com/IntelligenceModding/AdvancedPeripherals/pull/668#issuecomment-2498914233 (or a review on this PR)

zyxkad: well it is better to implement on 1.19.2, but it's not an mandatory requirement (so far?) if you don't know how to work on that. we can still backport the changes to 1.19.2 but it makes the merge a little bit harder.

zyxkad: what I mean for unexpected behaviour is if people are using discord bridge, when they set range to infinite they can bridge the chatbox message, while if they decreased the value, suddenly the bridge stop work for chatbox, which should not directly affect by the range setting.

Add a setting for this behaviour can clearly tells admins (in the comments) that bridge will only works if range is infinite. Such like the `chatBoxMultiDimensional` config.

Niki4tap: > well it is better to implement on 1.19.2, but it's not an mandatory requirement (so far?) if you don't know how to work on that. we can still backport the changes to 1.19.2 but it makes the merge a little bit harder.

Again, no problem with rebasing, this PR was mistakenly opened on the 1.20.x branch before I was aware of the guidelines, so I just want to make sure if it's wanted, and asking specifically from [SirEndii](https://github.com/SirEndii) since it looks like his review is required for the PR to get merged anyway.

If you have the access to merging and want to skip his review I can rebase and add a config option right away, no problem with that, just don't want to do work until I'm absolutely sure it's needed.

> what I mean for unexpected behaviour is if people are using discord bridge, when they set range to infinite they can bridge the chatbox message, while if they decreased the value, suddenly the bridge stop work for chatbox, which should not directly affect by the range setting.

See, for me it wouldn't be a surprise, if I want to send a "global" message, I would set the range to infinite and want ***everyone*** to see it. If it's a "local" message however, let's say I put 100 blocks radius, *just for my team* to see it, there is no reason I would want that to be bridged, since that would make the message "global" as before (the players won't see the message in minecraft chat, but they could just keep discord open or whatever platform the messages are being bridged to).

If I were to further clarify the behavior here, I would probably either mention it in the documentation of the chat box message APIs, or maybe create a different method, something like `broadcastFormattedMessage` instead of `sendFormattedMessage` with a clearer name and description, but I think that's a little too far and behavior should be intuitive, however if necessary I could go that route.

> Add a setting for this behaviour can clearly tells admins (in the comments) that bridge will only works if range is infinite. Such like the `chatBoxMultiDimensional` config.

The behavior shouldn't be clarified for the admins, they are not the ones using the peripherals, they should only be concerned with what players can do and what might be problematic. Configuration option just for a comment is the wrong place to tell information for the wrong people. The players need this information, and thus my aforementioned suggestion of clarifying it in the API documentation, not the configuration file, the players won't be looking there, and the admins don't need that information.

Niki4tap: Ah wait, nevermind, I think I misinterpreted "range" being the parameter to the API, instead of `maxRange` config setting, disregard the stuff I've said about the API docs then, and yeah the option would be fine.

Niki4tap:
Rebased and config option added. @zyxkad does this implementation look ok?

zyxkad (approved):

SirEndii:
@Niki4tap 
Thank you again from me for your nice work!

PR #668, opened 2024-11-21
Allow operation cooldown to be 0

Niki4tap:
- [ ] The commit message are well described (they are ok, can be amended if strictly necessary)

Allows for the cooldown to be set to 0, **the default cooldown is still the same**, so this doesn't change anything for most players, and allows for the folks who know what they're doing to lift the cooldown altogether.

Cooldown is set to at least 1000, and currently cannot be lower (unless you are going to patch the mod which I did in my survival world but wanted to upstream the changes)

Described above.

None, config change.

* **Other information**:
It's fine if this change cannot be upstreamed due to security concerns or whatever, I can still patch and build the mod myself for my worlds, but the cooldown can really get annoying if you're using the vanilla mod, and especially busy looping to wait for a message to leave the computer and sent in the chat makes the players want to code some sort of a "message sender domain" which is crazy as hell for this small thing, but is necessary for the correct and timely operation of computers, or if one trusts every computer on the server (you're playing only with a friend, not a public server), you could just lift the cooldown altogether and stop worrying about it.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/operations/SimpleFreeOperation.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/operations/SimpleFreeOperation.java
@@ -18,7 +18,7 @@ public enum SimpleFreeOperation implements IPeripheralOperation<Object> {
 
     @Override
     public void addToConfig(ForgeConfigSpec.Builder builder) {
-        cooldown = builder.defineInRange(settingsName() + "Cooldown", defaultCooldown, 1_000, Integer.MAX_VALUE);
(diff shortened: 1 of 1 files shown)


SirEndii:
Thank you for the contribution, but please keep in mind for the next time to make your changes for the oldest fully supported version as stated in the [guidelines](https://github.com/IntelligenceModding/AdvancedPeripherals/blob/dev/1.20.1/CONTRIBUTING.md#setting-up-a-development-environment)

Niki4tap:
Ah, sorry, do you mean targeting the PR for the older branch? Should I rebase #667 in that case?

SirEndii:
I am sorry that I didn't answer @Niki4tap 

It would be neat if you can rebase to 1.19.2/implement it in 1.19.2 due to my lack of time currently

Issue #660, opened 2024-09-30
getItemInHand() adding blank NBT to items


superjaja05:
### Describe

When using getItemInHand(), it seems to add some blank NBT to the held item, which causes them to be un-stackable with other regular items

### Steps to reproduce

- Link an inventory manager to yourself
- Hold an item that doesn't have nbt (like a dirt block)
- Use getItemInHand()

### Multiplayer?

Yes

### Version

1.20.1-0.7.40r (Latest 1.20.1)

### Minecraft, Forge and maybe other related mods versions

Minecraft 1.20.1, Forge 47.3.10

### Screenshots or Videos

https://github.com/user-attachments/assets/922def79-a7a5-4d86-a8fc-4790e348fcfe

PR #670, opened 2024-12-02
Merge dev/1.19.2 into dev/0.8

zyxkad:
- **make some change to fake player**
- **fix entity operation does not have enough reach range than expected**
- **Fixed Checkstyle violation**
- **[#660] Fixed blank nbt when using some inventory manager functions**
- **Removed redundant null check**
- **1.19.2-0.7.37r**
- **only clear owner when new empty card is swapped but not after load**
- **remove index check**

For further feature implementations


SirEndii:
Feel free to merge any dev/1.19.2 into dev/0.8 commits for yourself
I don't really need to review these

PR #675, opened 2024-12-21
Update build-and-test.yaml

zyxkad:
We should just use pull_request not pull_request_target

pull_request_target will executes code on the base commit, but not the PR commits.
pull_request_target is used for prevent secure hack, but I think we don't need worry about the hack?


zyxkad:
well okay, I need do more research to fix that

PR #682, opened 2025-01-11
release/lite 1.19.2

zyxkad:
Only kept vanilla integrations, chunky turtle & compass turtle


SirEndii:
I've started the game just to see what exactly is changed before I do a bigger code review (when I have more time)

So you changed the modid, but not the name, do we ant to do that?
There seems to be a second AP mod in the client but without an actual mod?

image


And how do we want to version this. Should we stay at 0.7.38, should it begin at 0.7.0 or do we want a completely different versioning? So we would start at 0.1 or maybe directly 1.0 since the lite version would not contain any of the 0.8 features or other stuff

SirEndii:
And do we maybe want to release a 1.18.2 version too?

zyxkad:
> There seems to be a second AP mod in the client but without an actual mod?

Yes, this is uses for register chunk ticket, and be able to migrate easily

zyxkad:
> And do we maybe want to release a 1.18.2 version too?

I wants to see if 1.19.2 works well, and if it is then I can port it to 1.16-1.18.

zyxkad:
> And how do we want to version this. Should we stay at 0.7.38, should it begin at 0.7.0 or do we want a completely different versioning? So we would start at 0.1 or maybe directly 1.0 since the lite version would not contain any of the 0.8 features or other stuff

Just for easy management I'd keep the version same, but only bump the version if main branch is updating on related logics. And player will get same result for same version of mod

Other mods should **not** dependent on the lite version mod, since its API may be removed/changed if its unnecessary.

SirEndii:
> > There seems to be a second AP mod in the client but without an actual mod?
> 
> Yes, this is uses for register chunk ticket, and be able to migrate easily

Alright I understand

src/main/java/de/srendi/advancedperipherals/AdvancedPeripherals.java
```diff
@@ -21,7 +19,8 @@
 @Mod(AdvancedPeripherals.MOD_ID)
 public class AdvancedPeripherals {
 
-    public static final String MOD_ID = "advancedperipherals";
+    public static final String MOD_ID = "advancedperipherals_lite";
+    public static final String ITEM_MOD_ID = "advancedperipherals";
     public static final String NAME = "Advanced Peripherals";
```
SirEndii, line 24: Do we also want to change this to `Advanced Peripherals Lite` so it stuff like our creative menu also have the new name

SirEndii: Maybe we also want to change the line 31 where we log that AP is loading to the lite name

SirEndii:
Question here
Iirc there was a reason why we didn't continue the lite version - what was that reason
And how would we handle the releases, I guess the standard workflow would be enough with some small changes and different project ids for modrinth and curseforge

I at least can't give this any love for now, my highest priority is 1.21 and 0.8

zyxkad:
> Iirc there was a reason why we didn't continue the lite version

Eh I thought you just reviewing it and forgot this XD. I don't think you said anything to reject this.

Yes standard workflow will be enough. In fact, I think this is rarely going to update since currently we are focusing on mod integrations, which don't affect this at all.

> I at least can't give this any love for now, my highest priority is 1.21 and 0.8

Sure I can try maintain this, it shouldn't be too much work for me atm.

SirEndii (approved):

src/main/java/de/srendi/advancedperipherals/AdvancedPeripherals.java
zyxkad: yeah sure, but you did not submit your review on January so I can't see them till now XD

PR #702, opened 2025-01-20
add CC Networking Cable p2p tunnel

zyxkad:

- [ ] Documentation, see https://github.com/IntelligenceModding/Advanced-Peripherals-Documentation/issues/96

  Feature

  Register AE2 p2p cable to forward cc:t connections

  No

* **Other information**:
   P2P tunnel item module is required
   
close IntelligenceModding/Advanced-Peripherals-Features#104


zyxkad:
I'm marking this as draft because it needs more test on edge cases.
@SirEndii You can review this if you want (and check why the item module is not rendering)

src/main/java/de/srendi/advancedperipherals/common/addons/ae2/WiredCableP2PTunnelPart.java
```diff
@@ -0,0 +1,142 @@
...
+            return Vec3.atCenterOf(WiredCableP2PTunnelPart.this.getBlockEntity().getBlockPos());
+        }
+
+        @Nonnull
+        @Override
+        public String getSenderID() {
+            return "p2p";
+        }
+
+        @Override
```
SquidDev, line 160: Hrm, how come you need to extend `WiredModemElement` (rather than `IWiredElement`) and implement `attachPeripheral` here? The tunnel neither provides or uses peripherals, so I'd expect this code to be redundant (though haven't tested).

zyxkad: Hmmm because I was following the cable's code, and need connect/disconnect the nodes when network changed.
I probably can just override `networkChanged` method

SirEndii:
So I tried to check how AE2 adds the item model
It does not seem that it tries to do that the same way it creates the block model. So I checked the resources from AE2 itself, there aren't any block models but the p2p item models. So I just created one in the following commit

src/main/java/de/srendi/advancedperipherals/common/data/ItemTagsProvider.java
```diff
@@ -19,5 +19,9 @@ protected ItemTagsProvider(DataGenerator generator, @Nullable ExistingFileHelper
     @Override
     protected void addTags() {
         tag(APTags.Items.SMART_GLASSES).add(APItems.SMART_GLASSES.get()).add(APItems.SMART_GLASSES_NETHERITE.get());
+        tag(de.srendi.advancedperipherals.common.addons.ae2.Integration.getCableP2PTag())
```
SirEndii: Is there a reason why you use the full qualifier here and don't import it?

SirEndii (changes requested):

src/main/java/de/srendi/advancedperipherals/common/data/ItemTagsProvider.java
zyxkad: > Is there a reason why you use the full qualifier here and don't import it?

Because `Integration` is defined in multiple packages. In order to not mix them up, I decide use full path here.

zyxkad:
Currently if you change the ME network (e.g. place one more cell/cable), existing peripherals will always detach and reattach.
I need to check if ME is reconstructing P2P parts when it's network changed, and if there are anyway to avoid the detach/attach

src/main/java/de/srendi/advancedperipherals/AdvancedPeripherals.java
```diff
@@ -67,4 +69,11 @@ public void commonSetup(FMLCommonSetupEvent event) {
         });
     }
 
+    public void onLoadComplete(FMLLoadCompleteEvent event) {
+        event.enqueueWork(() -> {
```
SirEndii, line 74: Do we maybe want to put this into the APAddons class or into the Integration class direclty

Do we even need a `Integration` class if the run method is not used? Maybe we want to re-organize that

SirEndii (changes requested):

src/main/java/de/srendi/advancedperipherals/AdvancedPeripherals.java
zyxkad: I was trying to do that in Integration.run(), but Integration runs before items get registered...
I'll move them to Registries

SirEndii: That is weird
Should happen after item registration
https://github.com/IntelligenceModding/AdvancedPeripherals/blob/016c6c2e3846b6732e7a5cd0afa7308abdf6cc91/src/main/java/de/srendi/advancedperipherals/common/setup/APRegistration.java#L48

SirEndii (approved):

PR #704, opened 2025-01-23
fix CraftJob return

zyxkad:

* **Other information**:

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/ae2/CraftJob.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/ae2/CraftJob.java
@@ -21,6 +21,7 @@ import org.jetbrains.annotations.Nullable;
 
 import java.util.concurrent.ExecutionException;
 import java.util.concurrent.Future;
+import java.util.concurrent.atomic.AtomicLong;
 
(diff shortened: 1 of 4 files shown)


src/main/java/de/srendi/advancedperipherals/common/addons/ae2/CraftJob.java
```diff
@@ -140,12 +142,9 @@ public void startCalculation() {
         if (startedCalculation) {
             return;
         }
+        startedCalculation = true;
 
         IGrid grid = node.getGrid();
-        if (grid == null) { //true when the block is not connected
```
zyxkad, line 145: AE2 source code prevent `getGrid` to return a `null`, I didn't check the version above 1.19.2 tho.
```java
    @Override
    public IGrid getGrid() {
        if (this.myGrid == null) {
            throw new IllegalStateException("A node is being used after it has been destroyed.");
        }
        return this.myGrid;
    }
```

SirEndii: 👍🏻

SirEndii:
Big thank you
This is still a prototype and I will maybe rewrite some of that stuff while abstracting all of this

PR #720, opened 2025-03-17
implement item&fluid handlers on bridges

zyxkad:
- **update gradle.properties**
- **implement ICapabilityProvider on MeBridgePeripheral close IntelligenceModding/Advanced-Peripherals-Features#77**
- **implement ICapabilityProvider on rsbridge**
- **make itemfilter default count as max integer so people can export/import all items when the count is not provided**

- [ ] Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/SirEndii/Advanced-Peripherals-Documentation/pulls). Feel free to remove this check if you don't need it

  Feature

  People can invoke `[export|import][Item|Fluid]` on other bridge peripherals, or invoke `[pull|push][Item|Fluid]` on bridges from normal inventories.

  The default item filter count changed from 64 to infinity (max integer)
  The default fluid filter count changed from 1000 to infinity (max integer)

* **Other information**:

### DIFF
--- a/gradle.properties
+++ b/gradle.properties
@@ -1,5 +1,5 @@
-org.gradle.jvmargs=-Xmx4G
 org.gradle.daemon=false
+org.gradle.jvmargs=-Xmx4G
 org.gradle.logging.level=info
 
 # Minecraft related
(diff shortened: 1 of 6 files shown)


SirEndii:
> The default filter count changed from 64 to infinity (max integer)

Why and shouldn't we also do that for the fluid filter then?

zyxkad:
> > The default filter count changed from 64 to infinity (max integer)
> 
> Why and shouldn't we also do that for the fluid filter then?

Yes I just forgot that

zyxkad:
> > The default filter count changed from 64 to infinity (max integer)
> 
> Why and shouldn't we also do that for the fluid filter then?

Because normally when player isn't providing any limit, it means no limit by intuition. Also 64 does not mean "a stack" for certain items such as tools, snowballs, etc.

SirEndii:
Good point

SirEndii:
Added it to the documentation issue https://github.com/IntelligenceModding/Advanced-Peripherals-Documentation/issues/96
*Maybe you want to document this*

PR #734, opened 2025-05-06
Add a global lua api to serve version helper functions

SirEndii:
Added a new global accessible lua api (to every computer, turtle and pocket) to serve helper functions to retrieve versions of AP, MC and our add-ons

I primarily had this idea when I made the new bridges and 0.8 to check if a script even supports your AP version.

image


src/main/java/de/srendi/advancedperipherals/common/addons/APAddons.java
```diff
@@ -1,12 +1,7 @@
 package de.srendi.advancedperipherals.common.addons;
 
-import de.srendi.advancedperipherals.AdvancedPeripherals;
-import net.neoforged.bus.api.SubscribeEvent;
 import net.neoforged.fml.ModList;
-import net.neoforged.fml.common.EventBusSubscriber;
-import net.neoforged.fml.event.lifecycle.InterModEnqueueEvent;
 
-@EventBusSubscriber(modid = AdvancedPeripherals.MOD_ID, bus = EventBusSubscriber.Bus.MOD)
 public class APAddons {
```
zyxkad: maybe we can just make this as enum instead?

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/luaapi/GlobalAPHelperAPI.java
```diff
@@ -0,0 +1,56 @@
...
+    public GlobalAPHelperAPI() {
+    }
+
+    @Override
+    public String[] getNames() {
+        return new String[]{NAME};
+    }
+
+    @LuaFunction
+    public final Object getVersions() {
```
zyxkad:
```suggestion
    public final Object getAddonVersions() {
```

src/main/java/de/srendi/advancedperipherals/common/addons/APAddons.java
SirEndii: Yeah that could be a good idea

SirEndii: I made an enum out of it, how I think it would be the best
What are your opinions of the approach I took?

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/luaapi/GlobalAPHelperAPI.java
```diff
@@ -0,0 +1,56 @@
...
+import net.neoforged.fml.ModContainer;
+import net.neoforged.fml.ModList;
+
+import java.util.HashMap;
+import java.util.Map;
+import java.util.Optional;
+
+public class GlobalAPHelperAPI implements ILuaAPI {
+
+    private static final String NAME = "ap_helper";
```
zyxkad: can we change `ap_helper` to something without `_` 😓 for example just `advancedperipherals`. or something shorter. Also the word `helper` does not really make sense for me.

SirEndii: I used an underscore since snake case is the default for peripheral names in CC
Not sure about api objects tho

Yeah, I will just rename it to `advancedperipherals` then

zyxkad: generally we should not use multiple words in library names. but `advancedperipherals` is a whole name and should be fine (maybe), as well as I cannot think of better name for it :<

src/main/java/de/srendi/advancedperipherals/common/setup/CCRegistration.java
```diff
@@ -50,7 +50,7 @@ public class CCRegistration {
 
     public static void register() {
         IntegrationPeripheralProvider.load();
-        ComputerCraftAPI.registerAPIFactory(unused -> new GlobalAPHelperAPI());
+        ComputerCraftAPI.registerAPIFactory(unused -> new GlobalVersionsLuaAPI());
```
zyxkad: since it's not related to computer environment, we can also use singleton for it.
```suggestion
        ComputerCraftAPI.registerAPIFactory(unused -> GlobalVersionsLuaAPI.INSTANCE);
```

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/luaapi/GlobalVersionsLuaAPI.java
```diff
@@ -13,12 +13,9 @@
 import java.util.Map;
 import java.util.Optional;
 
-public class GlobalAPHelperAPI implements ILuaAPI {
+public class GlobalVersionsLuaAPI implements ILuaAPI {
```
zyxkad, line 18:
```suggestion
    public static final String INSTANCE = new GlobalVersionsLuaAPI();

```

src/main/java/de/srendi/advancedperipherals/common/setup/CCRegistration.java
SirEndii: I agree

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/luaapi/GlobalVersionsLuaAPI.java
```diff
@@ -14,6 +14,7 @@
 import java.util.Optional;
 
 public class GlobalVersionsLuaAPI implements ILuaAPI {
+    public static final String INSTANCE = new GlobalVersionsLuaAPI();
```
SirEndii: why the hell is it a string

zyxkad (approved):

PR #735, opened 2025-05-06
Implement the new Bridge system for the ME Bridge

SirEndii:
A more descriptive description will come and a review is not yet recommend


src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AECraftJob.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+        AdvancedPeripherals.debug("Could not find CPU or job link even after job started", org.apache.logging.log4j.Level.WARN);
+    }
+
+    private CraftingJobStatus getJobStatus() {
+        if (jobStatus == null || jobStatus.get() == null && cachedStatus != null) {
+            return cachedStatus;
+        }
+        cachedStatus = jobStatus.get();
+        return jobStatus.get();
```
zyxkad:
```suggestion
        if (cachedStatus == null && jobStatus != null) {
            cachedStatus = jobStatus.get();
        }
        return cachedStatus;
```
I think you mean this

SirEndii: That makes more sense

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AECraftJob.java
```diff
@@ -0,0 +1,309 @@
...
+        }
+    }
+
+    private void prepareCPUAndStatus(ICraftingService service) {
+        if (jobLink == null || jobStatus != null || !startedCrafting) {
+            return;
+        }
+        for (ICraftingCPU cpu : service.getCpus()) {
+            if (cpu instanceof CraftingCPUCluster cpuCluster) {
+                if (cpuCluster.craftingLogic.getLastLink() != null && cpuCluster.craftingLogic.getLastLink().getCraftingID().equals(jobLink.getCraftingID())) {
```
zyxkad: Can we define a temporary variable here instead of copy paste same expr twice 😅

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AECraftJob.java
zyxkad: Actually I'm not sure if `jobStatus.get() == null` is on purpose, like if the cache should be used only after the job is finished or the status can be cached all the time?

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AECraftJob.java
SirEndii: Uh does that even work since the expression can change and the second expression is called later in the supplier?
Or do I overlook something here

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AECraftJob.java
SirEndii: I've added a comment which hopefully explains it a bit better

SirEndii:
@zyxkad Feature wise, well tested in my RS and AE2 world, everything should be pretty the same to RS in terms of properties of the objects and function syntax
The only thing that I am not too happy about is the layout off the AEApi file, but that's nothing I want to bother with currently (and don't have the time to). That's possibly a thing I want to change with 0.8? But tbh, not a priority even with the urgh code quality for me now
I don't see anything feature wise or maybe some potential performance issues that would need a change, but you have better eyes than me and I would love to release the beta of 1.21.1 this or next week

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AEApi.java
```diff
@@ -0,0 +1,935 @@
...
+
+    private static Class<? extends PatternContainer> tryCastMachineToContainer(Class<?> machineClass) {
+        if (PatternContainer.class.isAssignableFrom(machineClass))
+            return machineClass.asSubclass(PatternContainer.class);
+
+        return null;
+    }
+
+    public static <T extends AEKey> Map<String, Object> parseAeStack(Pair<Long, T> stack, @Nullable ICraftingService service) {
+        if (stack == null || stack.getRight() == null)
```
zyxkad: is stack becomes `null` expected? either way I'd recommend returns null instead of an empty map.

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AECraftJob.java
zyxkad: I mean
```suggestion
                final ICraftingLink lastLink = cpuCluster.craftingLogic.getLastLink();
                if (lastLink != null && lastLink.getCraftingID().equals(jobLink.getCraftingID())) {
```

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AEApi.java
```diff
@@ -0,0 +1,935 @@
...
+
+            InternalInventory inventory = entity.getInternalInventory();
+
+            for (int i = 0; i < inventory.size(); i++) {
+                ItemStack stack = inventory.getStackInSlot(i);
+
+                if (stack.getItem() instanceof ChemicalStorageCell) {
+                    BasicCellInventory cellInventory = BasicCellHandler.INSTANCE.getCellInventory(stack, null);
+
+                    used = cellInventory.getUsedBytes();
```
zyxkad:
```suggestion
                    used += cellInventory.getUsedBytes();
```

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AEApi.java
```diff
@@ -0,0 +1,935 @@
...
+            IItemHandler itemHandler = InventoryUtil.extractHandler(null, bus.getLevel(), connectedInventoryPos, bus.getSide());
+            if (itemHandler != null) {
+                for (int i = 0; i < itemHandler.getSlots(); i++) {
+                    ItemStack stack = itemHandler.getStackInSlot(i);
+
+                    // Use 64 as a rough estimate if no item is in that slot
+                    if (stack.isEmpty()) total += 64;
+
+                    // If the slot is not empty, use the max stack size of that stack
+                    total += stack.getMaxStackSize();
```
zyxkad:
```suggestion
                    total += stack.isEmpty() ? itemHandler.getSlotLimit(i) : stack.getMaxStackSize();
```

src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/MEChemicalHandler.java
```diff
@@ -0,0 +1,58 @@
...
+    @NotNull
+    @Override
+    public ChemicalStack insertChemical(int tank, ChemicalStack resource, @NotNull Action action) {
+        if(resource.isEmpty())
+            return resource;
+
+        ChemicalStack inserted = resource.copy();
+        long amountInserted = storageMonitor.insert(MekanismKey.of(resource), resource.getAmount(), action == Action.SIMULATE ? Actionable.SIMULATE : Actionable.MODULATE, actionSource);
+        inserted.setAmount(resource.getAmount() - amountInserted);
+        return inserted;
```
zyxkad:
```suggestion
        ChemicalStack remain = resource.copy();
        long amountInserted = storageMonitor.insert(MekanismKey.of(resource), resource.getAmount(), action == Action.SIMULATE ? Actionable.SIMULATE : Actionable.MODULATE, actionSource);
        remain.setAmount(resource.getAmount() - amountInserted);
        return remain;
```

zyxkad (changes requested):

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/MEBridgePeripheral.java
```diff
@@ -130,429 +155,796 @@ protected MethodResult importToME(@NotNull IArguments arguments, @Nullable IItem
...
+    @LuaFunction(mainThread = true)
+    public MethodResult getCraftingTask(int id) {
+        if (!isAvailable())
+            return notConnected(null);
+
+        AECraftJob foundJob = null;
+
+        for (AECraftJob job : bridge.getJobs()) {
+            if (job.getId() == id) {
+                foundJob = job;
```
zyxkad, line 867:
```suggestion
                foundJob = job;
                break;
```

zyxkad (approved):
otherwise lgtm

PR #737, opened 2025-05-18
New procedurally generated animations for the blocks

sjeulu:
Makes the parts of the blocks that were previously left static move, glow, shift, whatever you want to call it. The animations are generated by a tool that can be found [here](https://github.com/sjeulu/advanced-peripherals-util). It uses the old textures as the assets that it then modifies in order to build the animations, which is why the old textures are left untouched in the sources directory. This ensures that the original textures can be easily found and accessed by anyone who wishes to further improve or modify them, because applying changes to their post-processed versions can be somewhat tedious.


src/generated/resources/.cache/2f719b8031aed5e605ed8d6694b19245ca45c6ca
```diff
@@ -1,2 +1,7 @@
+<<<<<<< HEAD
```
zyxkad: guh conflicts :/

can you remove the `.cache` folder and do `runData` again?

sjeulu: done

zyxkad:
I noticed the highlight area changed, are they randomly generated? I personally think randomly generate stuff on this is not good, so maybe you can set a fixed seed?
Or idk what is @SirEndii 's opinion here.

sjeulu:
sure, i'll add an option to the script

sjeulu:
done

SirEndii:
Can you document the procedure exactly step by step [here](https://github.com/IntelligenceModding/AdvancedPeripherals/blob/dev/1.21.1/CONTRIBUTING.md), so I and other contributors know exactly how things will work, what we would need to do when we have new textures, trying to run the data gens etc

sjeulu:
sure, done

PR #757, opened 2025-06-30
add utf8 message support to chatboxes

456dev:
adds optional "useUtf8" parameter, informing us that the program is utf8 aware, providing its strings as utf8, This allows sending messages in the extended minecraft character set.

see the utf8 lib in cc, as well as the \u{xxxx} escape sequence.

also sends the events with a new utf8 message parameter.

- [X] The commit message are well described
- [ ] Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/IntelligenceModding/Advanced-Peripherals-Documentation/pulls). Feel free to remove this check if you don't need it (will todo tmr)
- [ ] All changes have fully been tested (TODO, test other functions, formatted/toast)

Feature

image

image

utf8 message is provided as argument 5 (the current last one)

NO: this behaviour is opt in as a default-false optional parameter to all messagebox methods.
if users with to take advantage of this behaviour when sending messages, they should set the (current) last parameter to true
unless they are relying on specific parameter counts / not having extra data in the event, but thats not a breaking change for 99%, just 1 case of [spacebar heating](https://xkcd.com/1172/)

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
@@ -172,43 +172,43 @@ public class ChatBoxPeripheral extends BasePeripheral<IPeripheralOwner> {
             String message = arguments.getString(0);
 
             // check size while it represents bytes (in utf8 mode) as that is longer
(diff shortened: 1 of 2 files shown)


SirEndii:
Greetings
Would you be able to port that to the dev/1.19.2 branch too?

Apart from that, I will try to review your PR asap

SirEndii (approved):

PR #758, opened 2025-07-01
backport port chatbox utf8 support to 1.19.2

456dev:
Backports #757 to 1.19.2

adds optional "useUtf8" parameter, informing us that the program is utf8 aware, providing its strings as utf8,
This allows sending messages in the extended minecraft character set.

see the utf8 lib in cc, as well as the \u{xxxx} escape sequence.

also sends the events with a new utf8 message parameter.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
@@ -24,7 +24,12 @@ import de.srendi.advancedperipherals.lib.peripherals.IPeripheralFunction;
 import de.srendi.advancedperipherals.network.APNetworking;
 import de.srendi.advancedperipherals.network.toclient.ToastToClientPacket;
 import net.minecraft.ChatFormatting;
-import net.minecraft.network.chat.*;
+import net.minecraft.network.chat.ClickEvent;
+import net.minecraft.network.chat.Component;
+import net.minecraft.network.chat.ComponentContents;
+import net.minecraft.network.chat.HoverEvent;
+import net.minecraft.network.chat.MutableComponent;
+import net.minecraft.network.chat.Style;
 import net.minecraft.resources.ResourceKey;
 import net.minecraft.server.level.ServerPlayer;
 import net.minecraft.world.item.TooltipFlag;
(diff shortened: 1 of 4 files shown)


src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
```diff
@@ -24,18 +24,12 @@
 import de.srendi.advancedperipherals.network.APNetworking;
 import de.srendi.advancedperipherals.network.toclient.ToastToClientPacket;
 import net.minecraft.ChatFormatting;
-import net.minecraft.network.chat.Component;
-import net.minecraft.network.chat.ComponentContents;
-import net.minecraft.network.chat.ClickEvent;
-import net.minecraft.network.chat.HoverEvent;
-import net.minecraft.network.chat.MutableComponent;
-import net.minecraft.network.chat.Style;
+import net.minecraft.network.chat.*;
```
zyxkad: Can we don't use import *? It is unclear for me.

zyxkad (changes requested):

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
SirEndii: yeah, checkstyle will also fail with these wildcard imports

456dev: I imported the checkstyle config into intellij as the code style, is there anything else i need to specificially modify? as-is checkstyle seems to pass

SirEndii: Interesting, I guess that rule was removed, or I hallucinate
Either way, please do not use wildcard imports

src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/ChatBoxPeripheral.java
```diff
@@ -349,11 +417,18 @@ public final MethodResult sendFormattedMessageToPlayer(@NotNull IArguments argum
     }
 
 
+    // 0 message, 1 title, 2 playerName, 3 prefix, 4 brackets, 5 bracket color, 6 range, 7 utf8compatible
```
zyxkad, line 420: In 0.8 we must use a keyword table argument instead of the long long argument list :(

SirEndii: I agree
Or at least only for the optional options

For 0.7, this is ok

PR #790, opened 2025-10-04
Add pattern grid turtles

JCLemme:
* **Please check if the PR fulfills these requirements**
- [X] The commit message are well described
- [X] Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/IntelligenceModding/Advanced-Peripherals-Documentation/pulls). Feel free to remove this check if you don't need it
- [~] All changes have fully been tested

* **What kind of change does this PR introduce?** (Bug fix, feature, ...)
Feature

* **What is the current behavior?** (You can also link to an open issue here)
Turtles cannot equip or otherwise interact with pattern grids

* **What is the new behavior (if this is a feature change)?**
Turtles can now equip pattern grids to write patterns: see below
https://github.com/IntelligenceModding/Advanced-Peripherals-Features/issues/125

* **Does this PR introduce a breaking change?** (What changes might users need to make in their scripts due to this PR?)
Not to my knowledge

* **Other information**:
I'd like to add unit tests if this project is set up for them

PR #791, opened 2025-10-14
Me Bridge Ae2 Pattern Creation

SchachSebastian:

- [ ] Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/IntelligenceModding/Advanced-Peripherals-Documentation/pulls). Feel free to remove this check if you don't need it
Not yet since some changes might be required

A new feature allowing one to create ae2 patterns using the me_bridge.
Only works if the connected me system has a pattern encoder.
Following new lua functions are added:
  - createCraftingPattern(resourceLocationString, allowSubstitutesBoolean, allowFluidSubstitutesBoolean)
creates all crafting patterns resulting in an item whose resourceLocation matches the given resourceLocationString (regex)

  - createSmeltingPattern(resourceLocationString)
creates all smelting patterns resulting in an item whose resourceLocation matches the given resourceLocationString (regex)

  - createBlastingPattern(resourceLocationString)
same as smelting, but for all blasting recipes

  - createSmokingPattern(resourceLocationString)
same as smelting, but for all smoking recipes

  - createSmithingPattern(resourceLocationString, allowSubstitutesBoolean)
creates all smithing patterns resulting  in an item whose resourceLocation matches the given resourceLocationString (regex)
does whoever currently ignore trim patterns
also this function requires reflection to obtain the inputs of the smithingRecipe

  - createStonecuttingPattern(resourceLocationString, allowSubstitutesBoolean)
creates all stonecutting patterns resulting  in an item whose resourceLocation matches the given resourceLocationString (regex)

and then one function that works a little differently

  - createProcessingPattern(inputs, outputs)
creates a processing pattern that takes the inputs specified in the inputs list and returns the outputs (the first output element will be considered as primary output)
parameter format:
```lua
-- inputs
{
  { "minecraft:diamond", 3 },
  { "minecraft:obsidian", 3 }
}
-- outputs
{
  { "minecraft:stone", 1 }
}
```

No, only new functions.

* **Other information**:
  - One might want to decide on only one way to create patterns (specifying the name of the output, or all the inputs and outputs). I'm leaving that to you, although I would prefer both :)
  - Currently, the creation of a pattern which result can be crafted in multiple ways will always produce multiple patterns. It would be possible to add an optional list of inputs that filters the recipes even further.
  - maybe renaming to match the rs variant ones, also not for me to decide
  - and maybe move it to dev/0.8 i'm not sure how your branches work

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/MEBridgePeripheral.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/computercraft/peripheral/MEBridgePeripheral.java
@@ -149,6 +149,13 @@ public class MEBridgePeripheral extends BasePeripheral<BlockEntityPeripheralOwne
 
         return MethodResult.of(FluidUtil.moveFluid(targetTank, fluidHandler, filter.getLeft()), null);
     }
+    /**
+     * removes blank patterns from the me system matching the number of given patterns and inserts the given patterns afterward.
+     * patterns that can not be inserted will be dropped at the location of the me bridge.
+     * @param patterns a list of ae2 patterns as itemStacks
+     * @return a {@link MethodResult}, NOT_FOUND if the given list is empty, MISSING_BLANK_PATTERN if there aren't enough blank patterns
+     * in the me system or PATTERN_CREATED if the process was successful.
+     */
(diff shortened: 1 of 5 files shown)


SirEndii:
I need to test this and #790 ingame 
Not a huge fan of two different ways to create patterns for each peripheral - but I need to test it to decide what we will do

PR #808, opened 2026-02-15
add nbtHash to itemstacks

Bubbet:

- [ ] Docs have been added / updated (for features or maybe bugs which were noted). If not, please update the needed documentation [here](https://github.com/IntelligenceModding/Advanced-Peripherals-Documentation/pulls). Feel free to remove this check if you don't need it

Feature

No CC:T nbt hash included, only a table of nbt tags

Added nbtHash, the hash CC:T would have made.

No

* **Other information**:
Not sure if fluids also have nbt, I only did the one area.

### DIFF
--- a/build.gradle
+++ b/build.gradle
@@ -232,16 +232,6 @@ repositories {
         }
     }
     maven {
-        url = uri("https://maven.pkg.github.com/refinedmods/refinedstorage")
-        credentials {
-            username = "anything"
-            password = "\u0067hp_oGjcDFCn8jeTzIj4Ke9pLoEVtpnZMP4VQgaX"
-        }
-        content {
-            includeModule("com.refinedmods", "refinedstorage")
-        }
-    }
-    maven {
         name = 'Kotlin for Forge'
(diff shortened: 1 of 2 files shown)


SirEndii:
Why do you remove cound and name?
And change the way the table is generated

Bubbet:
Those are included in that table that gets returned from what's inside the hashmap constructor. Tags are also included in those details. So that other method isn't necessary anymore?

Bubbet:
I only kept "maxStackSize" because its named differently, theres maxCount in that details too

SirEndii:
We can move the system in the future to use the DetailRegistry from CC in the right way but for now, I recommend to add it manually 

https://github.com/cc-tweaked/CC-Tweaked/blob/4bc04f14162aac62cb26dd6792fcd46413beb526/projects/common/src/main/java/dan200/computercraft/shared/details/ItemDetails.java#L32

zyxkad:
I'll close this, as we switched from fingerprint to nbthash in 0.8

PR #811, opened 2026-03-16
Fix: When installing this mod in the ATM10 V6.0 modpack, executing 's…

mcstomato:
* **请检查此 PR 是否符合以下要求 (Please check if the PR fulfills these requirements)**
- [x] 提交信息描述清晰 (The commit message are well described)
- [] 文档已添加/更新 (Docs have been added / updated)
- [x] 所有修改均已充分测试 (All changes have fully been tested)

* **此 PR 引入的变更类型 (What kind of change does this PR introduce?)**
Bug 修复 (Bug fix)

* **当前行为是什么？(What is the current behavior?)**
在 ATM10 V6.0 整合包中，当服务器后台 cmd 发送 say 且不带有参数（即仅发送 say）时触发服务器崩溃
When installing this mod in the ATM10 V6.0 modpack, executing 'say' from the server backend CMD without any parameters (i.e., just 'say') triggers a server crash

问题代码在 Events.java
The issue occurs in Events.java

根本原因：
Root cause:
MessageArgument.getMessage() 方法在参数不存在时会抛出 IllegalArgumentException
MessageArgument.getMessage() throws IllegalArgumentException when the parameter does not exist

旧代码没有处理这种异常情况
The previous code did not handle this exception

* **修复后的新行为是什么？(What is the new behavior?)**
添加了异常捕获，当参数不存在时不会导致服务器崩溃
Added exception handling to prevent server crash when the parameter does not exist

* **此 PR 是否引入破坏性变更？(Does this PR introduce a breaking change?)**
否
No

* **其他信息 (Other information)**
无
No

### DIFF
--- a/.vscode/launch.json
+++ b/.vscode/launch.json
@@ -1,0 +1,93 @@
+{
+  "version": "0.2.0",
+  "configurations": [
+    {
+      "type": "java",
+      "request": "launch",
+      "name": "Client",
+      "presentation": {
+        "group": "Mod Development - AdvancedPeripherals",
+        "order": 0
+      },
+      "projectName": "AdvancedPeripherals",
+      "mainClass": "net.neoforged.devlaunch.Main",
+      "args": [
+        "@E:\\Project_All\\Trae\\Advancedperipherals\\AdvancedPeripherals\\build\\moddev\\clientRunProgramArgs.txt"
+      ],
+      "vmArgs": [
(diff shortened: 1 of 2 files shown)


mcstomato:
Hi @zyxkad, just wanted to mention that the .vscode/launch.json file seems to be auto-generated (maybe from my local environment). I only modified Events.java. Should I remove it from the PR? Please let me know. Thanks!

zyxkad:
Thanks for your contribution. I've checked the code and noticed executing `/say` without any argument is in fact an invalid usage, so we will just ignore such case by checking `context.getCommand() == null `, instead of try-catch an exception.

PR #813, opened 2026-03-27
Fix AE2 Things mod ID

starcatmeow:

This PR fixes the mod ID of AE2 Things, it's `ae2things` not `ae2_things`. Ref: https://github.com/Technici4n/AE2Things-Forge/blob/4f2503cca033526a1299c800878360994d3649f1/src/main/java/io/github/projectet/ae2things/AE2Things.java#L38.

Currently if we have a 256k item cell from AE2 Things, it won't be shown in `getCells` call, due to the `APAddon.AE2_THINGS.isLoaded()` returns false even though it's loaded ([ref](https://github.com/IntelligenceModding/AdvancedPeripherals/blob/94850d8ad2416d39084dc1293f7ad9704bdb4081/src/main/java/de/srendi/advancedperipherals/common/addons/appliedenergistics/AEApi.java#L964)).

It will be correctly shown.

No.

### DIFF
--- a/src/main/java/de/srendi/advancedperipherals/common/addons/APAddon.java
+++ b/src/main/java/de/srendi/advancedperipherals/common/addons/APAddon.java
@@ -7,7 +7,7 @@ import java.util.Arrays;
 public enum APAddon {
 
     AE2("ae2"),
-    AE2_THINGS("ae2_things"),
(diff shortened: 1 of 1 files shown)


zyxkad:
thanks for the fix

PR #721, opened 2025-03-17
rename some variable names in InventoryUtil to make them have more sense

zyxkad:
**PLEASE READ THE [GUIDELINES](https://github.com/SirEndii/AdvancedPeripherals/blob/dev/1.20.1/CONTRIBUTING.md) BEFORE MAKING A CONTRIBUTION**

* **Please check if the PR fulfills these requirements**
- [x] The commit message are well described
- [x] All changes have fully been tested

* **What kind of change does this PR introduce?** (Bug fix, feature, ...)
  Code clean up

* **Does this PR introduce a breaking change?** (What changes might users need to make in their scripts due to this PR?)
  No

* **Other information**:
  I remember I did this one day, but clearly I forgot where I did that.

PR #817, opened 2026-04-10
Dev/0.8 bridges

zyxkad:
Don't know why this isn't merged to `dev/0.8`, we are losing some progress about bridges


SirEndii:
Iirc I manually merged/added every bridge feature to the 1.21.1 version since it was faster than a git merge

There could be differences between dev/1.21.1 and dev/0.8-1.21.1 tho, but dev/0.8-1.21.1 seems to be already up to date with all features
My knowledge could also not be up to date, I didn't work on that in a long while

zyxkad:
I found this out when realizing both #720 and ~~#721 (is not that important)~~ isn't applied for 1.21.1, I will merge them manually / implement again and then close this.

PR #836, opened 2026-09-12
Update ru_ru.json

Solar40990:
Update Russian localization for new content from 0.8.x


zyxkad:
Thank you for translation! However, you should target `dev/0.8-1.21.1` branch instead.

zyxkad:
ok a bit messy after change the target, but would you mind open a PR based on`dev/0.8-1.21.1` instead of `dev/1.21.1`?

ffgjjjwwww-stack:
> ok a bit messy after change the target, but would you mind open a PR based on`dev/0.8-1.21.1` instead of `dev/1.21.1`?

okay

